PR #40, opened 2024-10-22
Docs and tweaks

dvdplm:
Grab bag of minor changes that came up during code reading.

No major changes and fairly atomic commits, to allow for reviewers to reject parts as they please. :)


.gitignore
```diff
@@ -1 +1,4 @@
-/target
+target/*
```
fjarri, line 1: Just curious, why this change? It works the same, of course, but the previous variant makes the `target` folder in my editor's project view greyed out, while this one doesn't.

.gitignore
```diff
@@ -1 +1,4 @@
...
+Cargo.lock
```
fjarri, line 2: Note for the future: I wonder if we should actually commit it, and run tests against it (perhaps as a separate CI job), to distinguish cases where upstream broke something from our own bugs.

examples/README.md
```diff
@@ -1,5 +1,11 @@
+# Manul examples
+
 This crate illustrates the usage of `manul` for implementing distributed protocols.
 
-The library itself is the perspective of the protocol implementor, where they create a set of `Round` impls and write unit-tests for them.
+The library itself takes the perspective of the protocol implementor, as they create a set
```
fjarri, line 5: I'm not sure if it's some automatic on-save hook, but I intentionally don't wrap lines manually in .md files, since they're all text, and text editors manage that just fine themselves.

examples/tests/async_runner.rs
```diff
@@ -96,26 +112,34 @@ where
...
+            match session.preprocess_message(&mut accum, &incoming.from, incoming.message)? {
+                Some(preprocessed) => {
+                    // In production usage, this would happen in a separate task.
+                    debug!("{key:?}: Applying a message from {:?}", incoming.from);
+                    let processed = session.process_message(rng, preprocessed);
+                    // In production usage, this would be a host task.
+                    session.add_processed_message(&mut accum, processed)?;
+                }
+                None => {
+                    if accum.errors_present() {
```
fjarri, line 134: I don't think we should expose that. What would be the use of reporting the errors before finalizing the session? If it's only for the logging purposes, perhaps `preprocess_message()` can return a `Result` instead of `Option`.

fjarri: Well, `Result<Result<>>` that is, since we want to be able to immediately exit the loop on `LocalError`.

manul/src/protocol/round.rs
```diff
@@ -272,6 +272,7 @@ impl EchoBroadcast {
 }
 
 /// Message payload created in [`Round::receive_message`].
+#[derive(Debug)]
```
fjarri, line 275: Does it actually display the boxed object?

examples/tests/async_runner.rs
```diff
@@ -31,24 +33,38 @@ struct MessageIn<SP: SessionParameters> {
...
+/// Runs a session. Simulates what each participating party would run as the protocol progresses.
 async fn run_session<P, SP>(
     tx: mpsc::Sender<MessageOut<SP>>,
     rx: mpsc::Receiver<MessageIn<SP>>,
     session: Session<P, SP>,
 ) -> Result<SessionReport<P, SP>, LocalError>
 where
     P: 'static + Protocol,
-    SP: 'static + SessionParameters,
+    SP: 'static + SessionParameters + Debug,
```
fjarri, line 44: What causes the need for these bounds? This type doesn't get instantiated itself. If it's some other type is generic over it, we can use <https://docs.rs/derive-where/> to derive `Debug`/`Clone` for it - I already thought of doing that for some of the types.

examples/tests/async_runner.rs
```diff
@@ -31,24 +33,38 @@ struct MessageIn<SP: SessionParameters> {
...
 {
     let rng = &mut OsRng;
 
     let mut rx = rx;
 
     let mut session = session;
+    // Some rounds make use of messages from previous rounds. Such messages are stored here and
```
fjarri, line 51: I don't think that's a correct explanation. It's rather, some rounds can finalize earlier and send messages for the next round, which will be stored here until we start that round too.

examples/tests/async_runner.rs
```diff
@@ -219,20 +243,25 @@ where
 
 #[tokio::test]
 async fn async_run() {
+    // Instantiation of the protocol we're running.
```
fjarri, line 246: I think it's rather an alias, since it's still a type, not an instance.

manul/src/session/message.rs
```diff
@@ -158,7 +158,9 @@ impl<M> VerifiedMessage<M> {
     }
 }
 
-/// A message bundle to be sent to another node.
+/// A message bundle destined for another node.
```
fjarri, line 161: Btw, thinking of renaming it to just `Message`, since it's the one exposed to the public API, and the user doesn't need to worry what a "bundle" is.

manul/src/session/message.rs
```diff
@@ -261,7 +270,10 @@ where
         self.direct_message.payload()
     }
 
-    pub(crate) fn into_unverified(self) -> (Option<SignedMessage<EchoBroadcast>>, SignedMessage<DirectMessage>) {
+    /// Split the `VerifiedMessage` into its constituent parts: the [`DirectMessage`] and (possibly)
```
fjarri, line 273: `VerifiedMessageBundle`

manul/src/session/session.rs
```diff
@@ -11,7 +11,7 @@ use rand_core::CryptoRngCore;
 use serde::{Deserialize, Serialize};
 use serde_encoded_bytes::{Base64, SliceLike};
 use signature::{DigestVerifier, Keypair, RandomizedDigestSigner};
-use tracing::debug;
+use tracing::{debug, trace};
```
fjarri, line 14: So, when it comes to tracing, I am not sure how much we're supposed to log, and at what levels. So the traces currently in the code will be inconsistent, because I was just adding them for debugging.

manul/src/session/session.rs
```diff
@@ -503,6 +519,14 @@ where
         self.provable_errors.contains_key(from) || self.unprovable_errors.contains_key(from)
     }
 
+    /// Returns `true` is there are any errors stored.
+    ///
+    /// *Note*: Errors may be registered in a later round so this method returns the information we
+    /// have *thus far* and does not guarantee there will not be errors once the round is finalized.
+    pub fn errors_present(&self) -> bool {
```
fjarri, line 526: As I said above, I am not sure this method is particularly useful. The user might as well just wait for the `SessionReport`.

manul/src/session/session.rs
```diff
@@ -680,12 +704,12 @@ mod tests {
...
         // In order to support parallel message creation and processing we need that
-        // certain generic types could be Send and/or Sync.
+        // certain generic types be Send and/or Sync.
         //
         // Since they are generic, this depends on the exact type parameters supplied by the user,
-        // so if the user does not want parallelism, they may not use Send/Sync generic parameters.
-        // But we want to make sure that if the generic parameters are Send/Sync,
-        // our types are too.
+        // so if the user does not want parallelism, they can use generic parameters that are not
+        // Send/Sync generic parameters. But we want to make sure that if the generic parameters are
```
fjarri, line 711: The second "generic parameters" is probably a typo?

.gitignore
dvdplm: Oh I just copied the one from synedrion. Or was it crypto-bigint? I'll adjust!

.gitignore
dvdplm: Yeah I have never fully understood why the rule/advice is to never commit the lock file for libraries.

examples/README.md
dvdplm: It's a markdown linter doing this, yes. And I agree with you, it's annoying. I have tried a few different ones in the last few weeks but they all seem to insist on wrapping lines. Will try again.

examples/tests/async_runner.rs
dvdplm: Deriving `Debug` for the `RoundAccumulator` meant the `SP` had to be `Debug` as well. 

In general I think as many types as possible should derive `Debug`.

examples/tests/async_runner.rs
dvdplm: That's better wording for sure.

examples/tests/async_runner.rs
dvdplm: mmm, it's still a type for sure, but a generic type has one instantiation for each value of the generic parameter. So `Option<bool>` is a type instantiation of the type `Option<T>`. An "alias" is the actual same type with a different name, like `SmallPrime` in `crypto-primes` is an *actual* `u16`, not some generic instantiation of the type?

manul/src/session/message.rs
dvdplm: I agree! I wanted to suggest the same change.

manul/src/session/session.rs
dvdplm: I'll remove it, but here's the argument for something like this:

The `RoundAccumulator` is what is keeping the state as the protocol progresses. It is natural to expect a stateful object to answer queries about its current state. I do not think it's dangerous to do so and that users are comfortable with the idea of a stateful object changing state as the process progresses. In particular I think it's useful when designing a protocol.

manul/src/session/session.rs
dvdplm: ofc, ty.

manul/src/session/session.rs
dvdplm: I think we have seen from our Synedrion users that debugging these protocols is difficult. The overhead of logging is minimal when compared to multiple network rounds and bigint maths. I think we can be pretty liberal with logging. As for picking the "right" level, I think we can let that "self-configure" over time yeah? We do not need to invent strict rules about what is `DEBUG` and what is `TRACE`.

manul/src/protocol/round.rs
dvdplm: No, it prints `Payload(Any { .. })`

examples/tests/async_runner.rs
dvdplm: It's for logging yes. As mentioned elsewhere the control flow of `preprocess_message` is a bit unusual and surprising: a return value of `Ok(None)` can mean two things.

manul/src/session/message.rs
dvdplm: If `MessageBundle` becomes `Message` then logically `CheckedMessageBundle` ––> `CheckedMessage` and
`VerifiedMessageBundle` ––> `VerifiedMessage`, but there is already a `VerifiedMessage`, so that is
a bit awkward. :/
We can keep the `Bundle` for `Checked` and `Verified` messages ofc, but that's not super elegant either.

.gitignore
fjarri: Well, libraries are supposed to use whatever lock file the application depending on them uses, so they don't need one for production, but it is useful for tests to detect semver violations in dependencies. In other words, if you're adding `Cargo.lock`, it also needs a CI job that uses it. But I'm not quite sure whether it should be a new one (basically "stable tests" that won't get broken when ran a week after, and some dependency updated with a semver violation), or we can add it to an existing one.

examples/README.md
```diff
@@ -1,5 +1,9 @@
...
+
 This crate illustrates the usage of `manul` for implementing distributed protocols.
 
-The library itself is the perspective of the protocol implementor, where they create a set of `Round` impls and write unit-tests for them.
+The library itself takes the perspective of the protocol implementor, as they create a set of `Round` impls and write unit-tests for them.
+
+The integration tests are written from the perspective of the protocol user, emulating an asynchronous execution of the protocol on multiple nodes.
 
-The integration tests are the perspective of the protocol user, emulating an asynchronous execution of the protocol on multiple nodes.
+To run the example, execute: `RUST_LOG=debug cargo t -p manul-example --test async_runner`
```
fjarri, line 9: That reminds me, I renamed the folder from `example` to `examples`, but forgot to pluralize the crate as well.

manul/src/session/message.rs
fjarri: We can shorten the names of the user-facing types at the cost of lengthening the internal ones. So `VerifiedMessage` will become `VerifiedMessagePart` or something, and similarly for other related types.

examples/tests/async_runner.rs
fjarri: That's fair. Should we add `missing_debug_implementations` lint to the list in `lib.rs` then?

manul/src/session/session.rs
```diff
@@ -300,12 +314,13 @@ where
...
         } else if self.possible_next_rounds.contains(&message_round_id) {
             debug!(
-                "{:?}: caching message from {:?} for {:?}",
-                self.verifier(),
+                "{key:?}: Caching message from {:?} for {:?}",
                 verified_message.from(),
                 verified_message.metadata().round_id()
             );
             accum.cache_message(verified_message)?;
+            // TODO(dp): this is a bit awkward. It means "all good, but nothing to do here right
```
fjarri, line 322: Just spitballing here, what about returning a `Result<Preprocessed, LocalError>` where
```
enum Preprocessed {
    ToProcess(VerifiedMessageBundle<SP>),
    Cached,
    Error(...),
}
```
So that users could log these events if they wanted. In case they don't want to, it can have a method
```
impl Preprocessed {
    fn ok(self) -> Option<VerifiedMessageBundle<SP>> { ... }
}
```

dvdplm: Yes, having an `enum` to hold the three states is the way I think.

Maybe something for a different PR though?

examples/tests/async_runner.rs
dvdplm: Love it! Yes!

manul/src/session/message.rs
fjarri: Filed https://github.com/entropyxyz/manul/issues/42

manul/src/session/session.rs
fjarri: Filed https://github.com/entropyxyz/manul/issues/43

.gitignore
fjarri: Filed https://github.com/entropyxyz/manul/issues/44

Issue #25, opened 2024-10-13
Decouple `protocol` and `session` levels


fjarri:
Ideally we would want `protocol` to know nothing of `session`. But currently there are several back-references:
- `FirstRound::new()` needs to know about `SessionId` (see also #9),
- `LocalError` and `RemoteError` are used on both levels. Should we have each layer have their own?
- `EchoRoundError` is declared at the session level, but used at the protocol level as one of the variants of `ReceiveError`
- `Protocol` trait uses `Digest` trait, which is defined at the session level. Should we just move the whole `signing` module to the top level? Or separate `Digest` from the rest?

PR #41, opened 2024-10-22
Tie `SessionId` to `SessionParameters::Digest`

### LINKED ISSUE #9: `SessionId` improvements
- should we hash the incoming randomness? We will have to use `Protocol::Digest` for that. This will allow us to avoid using `Box`, but may create problems with `serde_bytes` (see #7)
- should `Round` layer know about `SessionId` (it is passed in `FirstRound::new()` as shared randomness), or should it just receive &[u8]? The latter will remove the annoying dependency of `protocol` layer on `session` layer.

fjarri:
Fixes #9
Affects #25 

- `SessionId` now hashes the incoming randomness using `SessionParameters::Digest` and keeps that hash. `SessionId::new()` renamed to `from_seed()`.
- `FirstRound::new` takes an `&[u8]` instead of a `SessionId` to reduce the coupling between layers.

This PR may possibly include making `SessionId` generic over `SessionParameters` and using `digest::Output<SP::Digest>` (which is a stack array) instead of `Box<[u8]>`, but that actually affects a lot of types (messages, evidences), so I am not sure if it's worth it.

### DIFF
--- a/examples/src/simple_malicious.rs
+++ b/examples/src/simple_malicious.rs
@@ -2,9 +2,7 @@ use alloc::collections::{BTreeMap, BTreeSet};
 use core::fmt::Debug;
 
 use manul::{
-    protocol::{
-        Artifact, DirectMessage, FinalizeError, FinalizeOutcome, FirstRound, LocalError, Payload, Round, SessionId,
-    },
+    protocol::{Artifact, DirectMessage, FinalizeError, FinalizeOutcome, FirstRound, LocalError, Payload, Round},
     session::signature::Keypair,
     testing::{round_override, run_sync, RoundOverride, RoundWrapper, TestSessionParams, TestSigner, TestVerifier},
 };
@@ -46,11 +44,11 @@ impl<Id: 'static + Debug + Clone + Ord + Send + Sync> FirstRound<Id> for Malicio
     type Inputs = MaliciousInputs<Id>;
     fn new(
         rng: &mut impl CryptoRngCore,
-        session_id: &SessionId,
+        shared_randomness: &[u8],
         id: Id,
         inputs: Self::Inputs,
     ) -> Result<Self, LocalError> {
-        let round = Round1::new(rng, session_id, id, inputs.inputs)?;
+        let round = Round1::new(rng, shared_randomness, id, inputs.inputs)?;
         Ok(Self {
             round,
             behavior: inputs.behavior,
--- a/manul/src/session/session.rs
+++ b/manul/src/session/session.rs
@@ -9,7 +9,7 @@ use core::fmt::Debug;
 use digest::Digest;
 use rand_core::CryptoRngCore;
 use serde::{Deserialize, Serialize};
-use serde_encoded_bytes::{Base64, SliceLike};
+use serde_encoded_bytes::{Hex, SliceLike};
 use signature::{DigestVerifier, Keypair, RandomizedDigestSigner};
 use tracing::{debug, trace};
 
@@ -52,7 +52,7 @@ pub trait SessionParameters {
 
 /// A session identifier shared between the parties.
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, PartialOrd, Ord, Hash)]
-pub struct SessionId(#[serde(with = "SliceLike::<Base64>")] Box<[u8]>);
+pub struct SessionId(#[serde(with = "SliceLike::<Hex>")] Box<[u8]>);
 
 /// A session ID.
 ///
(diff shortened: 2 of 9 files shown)


manul/src/session/session.rs
```diff
@@ -61,15 +61,24 @@ pub struct SessionId(#[serde(with = "SliceLike::<Base64>")] Box<[u8]>);
...
+    /// **Warning:** make sure the bytestring you provide will not be reused within your application.
+    /// Session ID collisions will affect error attribution and evidence verification.
+    pub fn from_seed<SP: SessionParameters>(bytes: &[u8]) -> Self {
+        Self(
+            SP::Digest::new_with_prefix(b"SessionId")
+                .chain_update(bytes)
+                .finalize()
+                .as_ref()
+                .into(),
+        )
```
dvdplm, line 81: Outside tests/benches/examples, why would we want users to create their own session identifiers? The safest way to use the API seems to not allow misuse ever, so: can we get away without this method completely?

If we cannot, is it possible to enforce some minimum length (of the seed)? Or is the `Digest` safe even with a single `0` as input?

manul/src/session/session.rs
```diff
@@ -61,15 +61,24 @@ pub struct SessionId(#[serde(with = "SliceLike::<Base64>")] Box<[u8]>);
 /// Must be created uniquely for each session execution, otherwise there is a danger of replay attacks.
 impl SessionId {
     /// Creates a random session identifier.
-    pub fn random(rng: &mut impl CryptoRngCore) -> Self {
-        let mut buffer = [0u8; 256];
+    pub fn random<SP: SessionParameters>(rng: &mut impl CryptoRngCore) -> Self {
```
dvdplm, line 64: In the same vein as the question below: why allow users to provide their own RNG? Outside tests/benchmarks, what is the use case for that?

Wouldn't users be safest if we a) picked a safe RNG and b) made this `random()` method the only way to generate a session id?

examples/src/simple_malicious.rs
```diff
@@ -45,11 +43,11 @@ impl<Id: 'static + Debug + Clone + Ord + Send + Sync> FirstRound<Id> for Malicio
     type Inputs = MaliciousInputs<Id>;
     fn new(
         rng: &mut impl CryptoRngCore,
-        session_id: &SessionId,
+        shared_randomness: &[u8],
```
dvdplm, line 46: "shared randomness" is a really good name for this. Maybe that's what we should rename `SessionId` to?

dvdplm, line 46: Can we do something clever here to enforce "good quality randomness"? Maybe a `static_assert` on the length at least?

manul/src/session/session.rs
fjarri: Do you propose that we create them randomly when a `Session` is created? I thought there may be use in making them deterministically based on e.g. current block hash and other network parameters.

dvdplm: Exactly, create them without any user provided input. Upstream should treat it as an opaque token, or "shared randomness" as it were. :)

fjarri: That's the thing though, how do you share it between the nodes? I don't see any other way but generating it based on some publicly available pseudo-random data.

manul/src/session/session.rs
fjarri: We don't have a system RNG in `no_std`, so that's really the only way to do it.

examples/src/simple_malicious.rs
fjarri: I'm not necessarily opposed to that. But even though it's the same object, it kind of serves different purposes on the session level and on the protocol level, so perhaps different names are justified?

fjarri: I don't know, I don't really want to police users like that. It's like those checks for the password length on websites, they only serve to annoy users.

fjarri: For example, the user of a `Session` will have to route messages to different sessions running concurrently, and it makes sense to route them based on something called "session id" rather than based on something called "shared randomness".

dvdplm: Expecting users to know what "good quality randomness" means is dangerous. At the very least we should provide advice in the docs, maybe something like "The shared randomness is a sort of nonce that should *never* be used more than once; the onus of ensuring this lies with the calling code. 32 random bytes provided by a quality CSPRNG is a good starting point."

manul/src/session/session.rs
dvdplm: Hmm, fair point. Maybe we can force users to implement the method themselves, like a "must-implement" trait? That way they can use the (hopefully) correct contextual information to ensure it never repeats. :/

manul/src/session/session.rs
dvdplm: https://github.com/rust-random/rand/blob/master/rand_chacha/README.md#crate-features is `no_std` though? Perhaps there are others too?

dvdplm (approved):
LGTM, leaving any outstanding tweaks to you, feel free to merge at will.

manul/src/session/session.rs
fjarri: Hm, I don't really see how that would be functionally different from the existing `from_seed` method. Let's get this PR in, and if you have ideas on how the process can be made safer, open an issue, and we'll explore that.

manul/src/session/session.rs
fjarri: Yes, but it has to be initialized with some randomness, and where will you get that from?

examples/src/simple_malicious.rs
fjarri: It's supposed to be used by developers of distributed systems, so hopefully they would have some idea about good cryptographic practices... But I don't really see a way of ensuring the randomness is good besides adding warnings to the docs.

fjarri:
Added more warnings to `SessionId::from_seed()` and exposed `SessionId::random()` for tests only. We can explore any improvements in follow-up PRs.

PR #45, opened 2024-10-24
CI improvements

### LINKED ISSUE #44: Add a "stable dependencies test" CI task
We need to add a test run in CI that uses `Cargo.lock` to be a reference point, with the fixed Rust version and dependencies, that will factor out possible semver violations.

fjarri:
- Various small CI tweaks (see separate commits)
- Add a "stable test" with a fixed Rust version and a lockfile. Fixes #44

Note: the docs for `cargo test` say about `--locked`: "Cargo will exit with an error when either of the following scenarios arises: ... Cargo attempted to change the lock file due to a different dependency resolution." I think it's not going to error out if some dependency bumps the patch version, but I'm not sure.

@dvdplm if you think that the stable test is not really needed, that can be reverted. I'm not sure myself.


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ env:
   RUSTFLAGS: "-Dwarnings"
 
 jobs:
-  build:
+  wasm-build:
```
dvdplm, line 15: Note to self: we should mention in the `README` that the crate is `nostd`, perhaps in the "goals" section.

.github/workflows/ci.yml
```diff
@@ -43,20 +43,22 @@ jobs:
...
+  # This is supposed to factor out possible bugs introduced by Rust compiler changes
+  # and dependency changes, making the results more reproducible.
+  stable-test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
-            rust: stable
+            rust: 1.81.0 # MSRV
```
dvdplm, line 61: I wish there was an "organization-wide" way of defining this, so that when we change it, it applies to all crates.

dvdplm (approved):

dvdplm:
As for the stable test: I have a (weak) preference for omitting the `--locked` because that means that the tests will indeed fail when deps go out of sync. I'd rather be alerted early to such things than tackling them all at once. That said, almost nobody agrees with me so I'm likely wrong!

Let's go with what the docs say for now and if it ever becomes annoying we can re-assess?

PR #60, opened 2024-11-02
Protocol combinators

fjarri:
We have several different situations where we need to build on top of an existing protocol or combine several of them:
1. Modifying messages the node sends, keeping `Round::Protocol` the same. This is useful for writing tests.
2. Chaining two protocols together. This is useful e.g. for joining Presigning and Signing in CGGMP. 
3. Executing two independent protocols simultaneously. This is useful e.g. for merging KeyInit and AuxGen in CGGMP. (They don't have to have the same number of round, but their execution graph has to be... compatible? Need to find a formal way to describe that).
4. Building on top of an existing protocol, adding features. This is useful e.g. for extending AuxGen to also do KeyRefresh in CGGMP. KeyRefresh needs some intermediate information created during AuxGen (Paillier keys), so the new protocol would have to introspect the AuxGen state, add information to the outgoing messages, extract a part of information from the received messages, and add logic to finalization too.

Status
- Scenario 1 done as `Misbehave`
- Scenario 2 done as `Chain`, but needs work on round nesting.

Notes:
- We need to avoid blanket implementations for traits, those can be painful on the users. Better create a newtype wrapper and impl stuff for it.
- All except for scenario 1 will involve creating a new `Protocol` implementation.
- It seems that scenario 4 is a superset of scenario 1, but the question is, if we implement it that way, will it add any boilerplate to the usage in tests? If it does, we better keep them separate.
- Scenario 2 has a problem: all the others do not need to modify round IDs, but this one does, since without it they will surely repeat. We need some mechanism of "nesting" round IDs. 
- Since the overriding methods have to introspect incoming messages, they need access to `Deserializer`. Should `ObjectSafeRound` `make_*` methods take the deserializer too?


examples/src/simple_chain.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl<'a, Id: PartyId> From<&'a NewInputs<Id>> for Inputs<Id> {
+    fn from(source: &'a NewInputs<Id>) -> Self {
+        source.0.clone()
+    }
+}
+
+impl<Id: PartyId> From<(NewInputs<Id>, u8)> for Inputs<Id> {
+    fn from(source: (NewInputs<Id>, u8)) -> Self {
+        source.0 .0
```
dvdplm, line 23:
```suggestion
        source.0.0
```

examples/src/simple_malicious.rs
```diff
@@ -19,132 +17,59 @@ use crate::simple::{Inputs, Round1, Round1Message, Round2, Round2Message};
...
-    inputs: Inputs<Id>,
-    behavior: Behavior,
-}
-
-#[derive(Debug)]
-struct MaliciousRound1<Id> {
-    round: Round1<Id>,
-    behavior: Behavior,
-}
+struct SimpleMaliciousProtocol;
```
dvdplm, line 25: In the same vein as the comment above: can we rename this to just `MaliciousRound`?

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,470 @@
...
+    vec::Vec,
+};
+use core::{fmt::Debug, marker::PhantomData};
+
+use rand_core::CryptoRngCore;
+use serde::{Deserialize, Serialize};
+
+use crate::protocol::*;
+
+/// A trait defining two protocols executed sequentially.
```
dvdplm, line 15: I haven't read the code yet but already here I wonder what guarantees we provide.
- Will Protocol2 run if Protocol1 fails?
- Will Proto2 receive as input the output of Proto1?
- If Proto1 succeeds and Proto2 fails, will the chained protocol fail?
  - …and in this case what happens to the successful output of Proto1 (e.g. artifacts, keys or whatever it produced)?

Do you think it's a good idea to document such matters here?

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,470 @@
...
+pub trait Chained<Id>: 'static
+where
+    Id: PartyId,
+{
+    /// The inputs of the new chained protocol.
+    type Inputs: Send + Sync + Debug;
```
dvdplm, line 21: Shouldn't this be the "sum" of the inputs of the two protocols? Or can it be something different?

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,470 @@
...
+#[serde(bound(deserialize = "
+    <<C::EntryPoint1 as EntryPoint<Id>>::Protocol as Protocol>::ProtocolError: for<'x> Deserialize<'x>,
+    <<C::EntryPoint2 as EntryPoint<Id>>::Protocol as Protocol>::ProtocolError: for<'x> Deserialize<'x>,
+"))]
+pub enum ChainedProtocolError<Id: PartyId, C: Chained<Id>> {
+    Protocol1(<<C::EntryPoint1 as EntryPoint<Id>>::Protocol as Protocol>::ProtocolError),
+    Protocol2(<<C::EntryPoint2 as EntryPoint<Id>>::Protocol as Protocol>::ProtocolError),
+}
+
+impl<Id, C> ChainedProtocolError<Id, C>
```
dvdplm, line 51: Could we do two `From` impls for this, one for each protocol error?

manul/src/combinators/misbehave.rs
```diff
@@ -0,0 +1,269 @@
...
+where
+    Id: PartyId,
+    B: Behavior,
+    M: Misbehaving<Id, B>,
+{
+    round: BoxedRound<Id, <M::EntryPoint as EntryPoint<Id>>::Protocol>,
+    behavior: Option<B>,
+}
+
+/// A trait defining a sequence of misbehaving rounds amending the messages sent by some existing ones.
```
dvdplm, line 33:
```suggestion
/// A trait defining a sequence of misbehaving rounds changing the messages sent by some existing, well behaved, rounds.
```

The word "amend" usually means to "correct" something or "make more accurate", so that clashes a bit with the misbehaviour context here. "Change" is pretty neutral; "manipulate" is malicious. Maybe there are more terms we could consider.

manul/src/combinators/misbehave.rs
```diff
@@ -0,0 +1,269 @@
...
+#[derive_where::derive_where(Debug; <M::EntryPoint as EntryPoint<Id>>::Inputs)]
+pub struct MisbehavingInputs<Id, B, M>
+where
+    Id: PartyId,
+    B: Behavior,
+    M: Misbehaving<Id, B>,
+{
+    /// The behavior for the rounds starting with these inputs.
+    ///
+    /// If `None`, all the amended behavior will be skipped.
```
dvdplm, line 105: Can you elaborate a bit on when one would want to skip the malicious behaviour? Like, why would I want to first setup a misbehaving protocol and then choose to skip the modified behaviour?

manul/src/protocol/errors.rs
```diff
@@ -68,6 +68,32 @@ impl<Id, P: Protocol> ReceiveError<Id, P> {
...
+    where
+        F: Fn(P::ProtocolError) -> T::ProtocolError,
+        T: Protocol,
+    {
+        ReceiveError(self.0.map::<T, F>(f))
+    }
+}
+
+impl<Id, P: Protocol> ReceiveErrorType<Id, P> {
+    pub(crate) fn map<T, F>(self, f: F) -> ReceiveErrorType<Id, T>
```
dvdplm, line 82:
```suggestion
    /// Maps `ProtocolError` from two different `Protocol` impls (`P` and `T`).
    pub(crate) fn map<T, F>(self, f: F) -> ReceiveErrorType<Id, T>
```

manul/src/protocol/object_safe.rs
```diff
@@ -192,38 +199,86 @@ where
...
             .map_err(|_| LocalError::new(format!("Failed to downcast into type {}", core::any::type_name::<T>())))
     }
+
+    /// Attempts to provide a reference to an object of a concrete type.
+    ///
+    /// Fails if the wrapped type is not `T`.
+    pub fn downcast_ref<T: Round<Id>>(&self) -> Result<&T, LocalError> {
+        if self.wrapped && self.boxed_type_is::<ObjectSafeRoundWrapper<Id, T>>() {
+            let ptr: *const dyn ObjectSafeRound<Id, Protocol = P> = self.round.as_ref();
+            // This should be safe since we just checked that we are casting to a correct type.
```
dvdplm, line 270:
```suggestion
            // Safety: This is safe since we just checked that we are casting to the correct type.
```

manul/src/protocol/round.rs
```diff
@@ -2,89 +2,107 @@ use alloc::{
...
+impl Display for RoundId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        /*write!(f, "Round ")?;
+        for i in (0..self.depth as usize).rev() {
+            write!(f, "{}", self.round_num.get(i).ok_or_else(fmt::Error::)?)?;
+            if i != 0 {
+                write!(f, "-")?;
+            }
+        }
+        Ok(())*/
```
dvdplm, line 49: cleanup?

manul/src/protocol/round.rs
```diff
@@ -2,89 +2,107 @@ use alloc::{
...
+    pub const fn new(round_num: u8) -> Self {
         Self {
-            round_num,
+            depth: 1,
+            round_num: [round_num, 0, 0, 0, 0, 0, 0, 0],
             is_echo: false,
         }
     }
 
+    pub(crate) const fn group_under(&self, round_num: u8) -> Self {
```
dvdplm, line 73: Do you think we need more docs to describe the purpose and behaviour of this (i.e. panicking behaviour).

manul/src/protocol/round.rs
```diff
@@ -244,6 +264,40 @@ pub trait ProtocolError: Debug + Clone + Send {
...
+        panic!("Attempt to use an empty error type in an evidence. This is a bug in the protocol implementation.")
+    }
+}
+
+/// Describes unattributable errors originating during protocol execution.
+///
+/// In the situations where no specific message can be blamed for an error,
+/// each node must generate a correctness proof proving that they performed their duties correctly,
+/// and the collection of proofs is verified by a third party.
+/// One of the proofs will be necessarily missing or invalid.
```
dvdplm, line 294:
```suggestion
/// One of the proofs will necessarily be missing or invalid.
```

CHANGELOG.md
```diff
@@ -31,6 +33,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Re-export `digest` from the `session` module. ([#56])
 - Added `Message::destination()`. ([#56])
 - `PartyId` trait alias for the combination of bounds needed for a party identifier. ([#59])
+- An impl of `ProtocolError` for `()` for protocols that don't use errors. ([#60])
+- A dummy `CorrectnessProof` trait. ([#60])
+- A `misbehave` combinator, intended primarily for testing. ([#60])
+- A `chain` combinator for chaining two protocols. ([#60])
+- `EntryPoint::ENTRY_ROUND` constant. ([#60])
```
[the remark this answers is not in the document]
fjarri: Glad to hear that, that's what I always do in all my projects :) Usually it's only one line per PR because the PR only contains one public change, but this one is pretty large.

examples/src/simple.rs
```diff
@@ -149,14 +149,15 @@ struct Round1Payload {
     x: u8,
 }
 
-impl<Id: PartyId> FirstRound<Id> for Round1<Id> {
+impl<Id: PartyId> EntryPoint<Id> for Round1<Id> {
```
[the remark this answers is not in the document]
fjarri: It was logical when `FirstRound` was bound on `Round`, but it's not anymore. Now it's more like a Round factory.

examples/src/simple_chain.rs
```diff
@@ -0,0 +1,84 @@
+use core::fmt::Debug;
+
+use manul::{
+    combinators::{ChainEntryPoint, Chained},
+    protocol::PartyId,
+};
+
+use super::simple::{Inputs, Round1};
+
+pub struct ChainedSimple;
```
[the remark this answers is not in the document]
fjarri: It just refers to the `SimpleProtocol`. Could very well be `FooProtocol`, maybe it should be. Or, ideally, some simple but actually meaningful protocol that we could use for illustration. Simplified FROST, maybe?

examples/src/simple_chain.rs
fjarri: That's what `rustfmt` does. I suppose I can unstructure the tuple into variables first.

examples/src/simple_malicious.rs
fjarri: I wanted it to refer to the `SimpleProtocol`, so perhaps `MaliciousSimpleProtocol` would be more accurate.

manul/src/combinators/chain.rs
fjarri: > Will Protocol2 run if Protocol1 fails?

No

> If Proto1 succeeds and Proto2 fails, will the chained protocol fail?

Yes

I thought that it's the logical expectation, it basically makes them into one protocol, with the same guarantees as any other protocol. But perhaps it should be spelled out explicitly. 

> Will Proto2 receive as input the output of Proto1?
> …and in this case what happens to the successful output of Proto1 (e.g. artifacts, keys or whatever it produced)?

That's what the bound
```rust
type EntryPoint2: EntryPoint<
        Id,
        Inputs: From<(
            Self::Inputs,
            <<Self::EntryPoint1 as EntryPoint<Id>>::Protocol as Protocol>::Result,
        )>,
    >;
```
is for.

I already used it in the https://github.com/entropyxyz/synedrion/pull/156 to make InteractiveSigning, if you want to see an example.

manul/src/combinators/chain.rs
fjarri: It is, but I didn't want to just use a tuple because inputs may be intersecting (e.g. in InteractiveSigning both constituent protocols take `KeyShare` and `AuxInfo`)

manul/src/combinators/chain.rs
fjarri: I think that was the original variant, but I had some problems with it. I'll try it again, don't remember what the problems were. But since it's private, explicit conversion has its benefits.

manul/src/combinators/misbehave.rs
fjarri: You would usually want one round to be malicious and the rest normal. But `Session::new()` only takes one entry point, so it needs to be able to generate both types.

manul/src/protocol/round.rs
fjarri: Oh yes, I was deciding between the dumb procedural approach and the functional one, and forgot to delete this one.

manul/src/protocol/round.rs
fjarri: More docs are always welcome. I am not even sure about the name, feels like there should be a better term.

CHANGELOG.md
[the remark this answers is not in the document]
fjarri: Also note that each of these lines is added in the corresponding commit that introduces the change - easier to locate the change later. That's why I like doing merge commits in PRs.

examples/src/simple.rs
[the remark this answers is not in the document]
fjarri: This will be especially true if we make it stateful (see the note in the PR description).

examples/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 extern crate alloc;
 
 pub mod simple;
+pub mod simple_chain;
```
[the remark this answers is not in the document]
fjarri: We do need a lot more documentation in the examples, I've been just delaying that because I wanted to push through to have all the stuff I need for https://github.com/entropyxyz/synedrion/pull/156. Filed https://github.com/entropyxyz/manul/issues/63, I propose to do that in a separate PR.

examples/src/simple_chain.rs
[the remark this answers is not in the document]
fjarri: I think this falls under #63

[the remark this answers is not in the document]
fjarri: > How about we just call this `Chained`? Or `ChainedProtocol`?

Both of these are already exported from `combinators::chain`

examples/src/simple_chain.rs
fjarri: Fixed.

examples/src/simple_malicious.rs
fjarri: But you are right in a sense that it is not actually a protocol (that is, not something implementing `Protocol`), but rather a token type for which a specific misbehavior is defined.

fjarri: Renamed to `MaliciousLogic`.

manul/src/combinators/chain.rs
fjarri: Tried that, and found out what the problem was: it conflicts with the generic `impl<T> From<T> for T` blanket impl in the standard library.

manul/src/combinators/misbehave.rs
fjarri: Changed to "modify", but that's neutral. Perhaps we should pick something with an "evil" connotation.

manul/src/protocol/round.rs
fjarri: Fixed. Actually I'll keep the procedural approach, the functional one is way too elaborate.

manul/src/protocol/round.rs
fjarri: Added some docstrings.

manul/src/combinators/chain.rs
dvdplm: My (poorly worded) comment was not so much questions I had as suggestions for things to cover in the docs. I do think the guarantees/behaviour/expectations should be spelled out explicitly.

One more thing came to mind today: side effects. If proto1 succeeds and proto2 fails, the sum fails. But there are no facilities to "roll back" any side effects. Authors must be aware that proto2 does not somehow "call back into proto1" to clean up or roll back any changes made there.

manul/src/combinators/chain.rs
dvdplm: Hmm, so that's a sort of "implicit invariance expectation" we have. I don't think there's any way to express this in the type system ("`Input` is a de-duped sum of proto1 and proto2's `Input`s"). :/

Good docs needed!

manul/src/combinators/chain.rs
dvdplm: Got it, thank you for checking.

manul/src/combinators/chain.rs
fjarri: Not sure what side effects you mean here. By changes, do you mean the changes in the accumulator? Everything else is immutable.

manul/src/combinators/chain.rs
fjarri: Well, the `From` bound on `EntryPoint2::Inputs` is the way to express it in the type system, although it does take some effort to parse it mentally. Another option is to introduce some kind of a `Split` trait that would generate `(EntryPoint1::Inputs, EntryPoint2::Inputs)`.

manul/src/combinators/chain.rs
fjarri: Added some documentation.

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,520 @@
...
+For all external appearances the new protocol is a single entity with its own
+[`Protocol`](`crate::protocol::Protocol`) type and an [`EntryPoint`](`crate::protocol::EntryPoint`) type.
+
+For example, imagine we have a `ProtocolA` with an entry point `EntryPointA`, inputs `InputsA`,
+two rounds, `RA1` and `RA2`, and the result `ResultA`;
+and similarly a `ProtocolB` with an entry point `EntryPointB`, inputs `InputsB`,
+two rounds, `RB1` and `RB2`, and the result `ResultB`.
+
+Then the chained protocol will provide `ProtocolC: Protocol` and `EntryPointC: EntryPoint`,
+the user will define `InpputsC` for the new protocol, and the execution will look like:
```
dvdplm, line 15:
```suggestion
the user will define `InputsC` for the new protocol, and behind the scenes the execution will look like:
```

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,520 @@
...
+- `InputsA` created from `InputsC` via the user-defined `From` impl;
```
dvdplm, line 16: Users providing the `From` impl clashes a little bit with the claim that "for all external observers the new protocol is a single entity": users do need to know enough about the internal protocols to write the `From` impl.
This doesn't seem like a big blocker or even a bad thing, but perhaps it's good to tweak the wording a bit?

dvdplm: As a style nitpick: I think switching to present tense would make the text easier to read ("`RA1` executes…", "`InputsB` is created").

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,520 @@
...
+- `EntryPointA` initialized with `InputsA`;
+- `RA1` executed;
+- `RA2` executed, producing `ResultA`;
+- `InputsB` created from `ResultA` and`InputsC` via the user-defined `From` impl;
+- `RB1` executed;
+- `RB2` executed, producing `ResultB`.
```
dvdplm, line 22: Is it worth noting that `ResultB` is the result for the chained protocol you think?

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,520 @@
...
+
+If the execution happens in a [`Session`](`crate::session::Session`), and there is an error at any point,
+a regular evidence or correctness proof are created using the corresponding types from the new `ProtocolC`.
```
dvdplm, line 25: But errors in `ProtocolA` will use the types of `ProtocolA` right? Are they wrapped in `ProtocolC`s types then?

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,520 @@
...
+
+The usage is as follows.
+
+1. Define a type that will serve as the inputs for the new joined protocol.
+   Most likely it will be a union between inputs of the first and the second protocol.
+
+2. Implement [`Chained`] for a type of your choice. Usually it will be an empty token type.
```
dvdplm, line 32: "empty token type": you mean like a ZST e.g. `struct ProtoC`?

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,520 @@
...
+   You will have to specify the entry points of the two protocols,
+   and the [`From`] conversions from the new input type to the inputs of both entry points
+   (see the corresponding associated type bounds).
+
+3. The entry point for the new protocol will be [`ChainedEntryPoint`] parametrized with
+   the type implementing [`Chained`] from the step 2.
```
dvdplm, line 38:
```suggestion
   the type implementing [`Chained`] from step 2.
```

manul/src/combinators/chain.rs
```diff
@@ -0,0 +1,520 @@
...
+- `InputsB` created from `ResultA` and`InputsC` via the user-defined `From` impl;
+- `RB1` executed;
+- `RB2` executed, producing `ResultB`.
+
+If the execution happens in a [`Session`](`crate::session::Session`), and there is an error at any point,
+a regular evidence or correctness proof are created using the corresponding types from the new `ProtocolC`.
+
+The usage is as follows.
+
+1. Define a type that will serve as the inputs for the new joined protocol.
```
dvdplm, line 29: Consider switching to present tense for readability, so "a type that serves…", "Most likely it is the union…" etc.

manul/src/combinators/misbehave.rs
```diff
@@ -0,0 +1,291 @@
...
+
+use crate::protocol::{
+    Artifact, BoxedRng, BoxedRound, Deserializer, DirectMessage, EchoBroadcast, EntryPoint, FinalizeError,
+    FinalizeOutcome, LocalError, NormalBroadcast, ObjectSafeRound, PartyId, Payload, ReceiveError, RoundId, Serializer,
+};
+
+/// A trait describing required properties for a behavior type.
+pub trait Behavior: 'static + Debug + Send + Sync {}
+
+impl<T: 'static + Debug + Send + Sync> Behavior for T {}
```
dvdplm, line 39: Maybe this belongs one level up, in `combinators.rs`, given it's more general than just misbehaviour?

manul/src/combinators/misbehave.rs
```diff
@@ -0,0 +1,291 @@
...
+    B: Behavior,
+    M: Misbehaving<Id, B>,
+{
+    round: BoxedRound<Id, <M::EntryPoint as EntryPoint<Id>>::Protocol>,
+    behavior: Option<B>,
+}
+
+/// A trait defining a sequence of misbehaving rounds modifying or replacing the messages sent by some existing ones.
+///
+/// Override one or more optional methods to modify the specific messages.
```
dvdplm, line 57: Just thinking out loud: I wonder if as time goes by people will start using the facilities in this module to do stuff that is not strictly "misbehavior". E.g. like the [`inspect`](https://doc.rust-lang.org/std/iter/struct.Inspect.html) combinator. It acts almost like a sort of middle ware.

manul/src/combinators/chain.rs
fjarri: But that's the same for a regular protocol. `manul` just calls the hooks and gives them data, what happens in the hooks is user's business.

manul/src/combinators/chain.rs
fjarri: > Users providing the From impl clashes a little bit with the claim that "for all external observers the new protocol is a single entity"

I guess I'm mixing two hypothetical users here: the one who writes the protocol, and the one who uses it. The latter doesn't need to care about the `From` implementations.

manul/src/combinators/misbehave.rs
fjarri: See https://github.com/entropyxyz/manul/issues/62. Maybe we can make it more general, or maybe it'll make writing tests harder, then we need a separate one.

manul/src/combinators/chain.rs
fjarri: Rephrased the statement about the appearance of the protocol, and added verbs to the execution list (technically it was already in the present tense, just with "to be" omitted)

manul/src/combinators/chain.rs
fjarri: They are, but neither the protocol writers nor the protocol runners need to worry about that, the routing happens automatically.

manul/src/combinators/chain.rs
fjarri: Yep. Do you think "zero-sized type" would be more understandable here? Technically, it doesn't have to be zero-sized, it's just never instantiated by the crate.

manul/src/combinators/chain.rs
fjarri: I am not sure about it, the instructions here are of the form "if you do this, that will happen". If you write "Define a type that serves as the inputs ...", doesn't that imply that this type somehow already has this property?

manul/src/combinators/misbehave.rs
fjarri: It is used only by `misbehave`, and probably won't be used by anything else. It's not even a meaningful trait, just an alias so that I don't have to repeat all the bounds.

manul/src/protocol/round.rs
```diff
@@ -5,86 +5,113 @@ use alloc::{
...
+// We need this to be limited to allow the nesting to be performed in `const` context
+// (since we cannot use heap there).
+const ROUND_ID_DEPTH: usize = 8;
 
 /// A round identifier.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct RoundId {
-    round_num: u8,
+    depth: u8,
+    round_nums: [u8; ROUND_ID_DEPTH],
```
fjarri, line 41: Actually thinking of switching from `EntryPoint::ENTRY_ROUND` to `EntryPoint::entry_round()`, this way we won't need constant methods on `RoundId`, and can use `Vec` without bothering about depth limits.

fjarri: Did that, using `smallvec` so in the majority of cases there won't be any allocations.

manul/src/combinators/chain.rs
dvdplm: I think "token" is the correct term to use here, but given the world we live in, I suspect readers may confuse "token" with "cryptocurrency token".
Given you have a "usually" here, it doesn't have to be technically exact, it's more of a suggestion.

How about "Usually it will be a (possibly zero-sized) marker type."? Otherwise "token type" is fine too.

manul/src/combinators/chain.rs
dvdplm: No, not necessarily (but this is subtle and I'm not a native speaker so not 100% sure).

How about this: "1. Define an input type for the new joined protocol"?

manul/src/combinators/misbehave.rs
dvdplm: Fair!

manul/src/combinators/misbehave.rs
dvdplm: okie, we can come back and tweak it if it ever become cumbersome.

dvdplm:
> * Naming is kind of awkward. Suggestions?

I'm struggling with this. I don't love `EntryPoint` but it's pretty clear what it is/does. Let's go with that and if either of us have a brilliant idea we'll revisit.

> 
> * Random idea: should `EntryPoint::new` take `&self`? This way we can basically formalize the concept of protocol constructors used in `synedrion`.

I like this! 

> 
> * Should `misbehave` go into `testing`? Or gated behind the `testing` feature?

I like `misbehave` to be difficult to misuse. Today we want to use it in tests, so let's put it there until we need it. I can totally envision a scenario where the core team wants to build a malicious node to test on (private?) testnets so maybe it'd be useful to have `misbehave` under a feature, but let's do that when the need arises.

> * The system with `From` bounds in the chain combinator is a little awkward, since we have to clone stuff for the first protocol. Some kind of a "splitting" trait might be more idiomatic, to split the full inputs into the inputs for Protocol 1 and the inputs for Protocol 2.
> 
> * Also it requires us to impl `From` for the type from the original protocol, which may be foreign in practice.

Hmm. I like the idea of a "splitting" trait, but I don't have a strong opinion on this.

dvdplm (approved):
This LGTM. I left a few more thoughts&nitpicks – feel free to make the changes or not and then merge.

fjarri:
> Random idea: should EntryPoint::new take &self? This way we can basically formalize the concept of protocol constructors used in synedrion.
> Hmm. I like the idea of a "splitting" trait, but I don't have a strong opinion on this.

This is actually coming in #68

fjarri:
I remember seeing your comment about `tinyvec` vs `smallvec`, and, having looked at them, I agree that `tinyvec` is preferable.

PR #67, opened 2024-11-07
Logic fixes for rounds with partial echo

### LINKED ISSUE #21: Add an example of a protocol that only sends messages to a subset of nodes
This will be similar to a key resharing protocol. We have a set of "old holders" that sends messages to a set of "new holders", and these sets may intersect or coincide.

fjarri:
Fixes #21

Imagine the following round configuration (which is possible during KeyResharing from `synedrion`):
- node 0 sends an echo-broadcast to [1, 2, 3]
- node 1 sends an echo-broadcast to [2, 3]
- nodes 2 and 3 do not send echo-broadcasts

After we received the initial broadcasts, we want an echo round where the nodes 1, 2, 3 echo received messages to each other. That is:
- node 0 does not send or receive anything
- node 1 sends the broadcasts from [0] to 2 and 3, and expects [0, 1] from 2 and 3
- node 2 sends the broadcasts from [0, 1] to 1 and 3, and expects [0] from 1 and [0, 1] from 3
- node 3 sends the broadcasts from [0, 1] to 1 and 2, and expects [0] from 1 and [0, 1] from 2

(note that node 1 does not send its broadcast the second time; that would be pointless)

This requires adding another method to `Round` specifying the exact behavior of the node, since this information cannot be obtained from existing methods. This PR introduces `Round::echo_round_participation()` and `EchoRoundParticipation` enum. 

In most cases we can use the blanket implementation returning `EchoRoundParticipation`, which means that the node either sends and receives echo messages, or does neither. This means that we can deduce the properties of the echo round from `Round::message_destinations()` and `Round::expecting_messages_from()`. The case where the node only sends echo messages (node 0), we don't need any additional info other than this fact. For the case where the node only receives echo messages (node 3), it needs additional info specifying which nodes actually participate in the echo round.

*Edit:* another possible API here is to get rid of `EchoRoundParticipation` enum and instead have a method returning `Option<BTreeSet<Id>>`. `None` would cover both `Default` and `Send`, and we will distinguish between them by checking if `expecting_messages_from()` is empty or not. Would that work?

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,1 +1,0 @@
-allow-unwrap-in-tests = true
--- a/manul/Cargo.toml
+++ b/manul/Cargo.toml
@@ -29,12 +29,14 @@ serde_json = { version = "1", default-features = false, features = ["alloc"], op
 
 [dev-dependencies]
 impls = "1"
+rand_core = { version = "0.6.4", default-features = false, features = ["getrandom"] }
 rand = { version = "0.8", default-features = false }
 serde_asn1_der = "0.8"
 criterion = "0.5"
 serde-persistent-deserializer = "0.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
 serde_json = { version = "1", default-features = false, features = ["alloc"] }
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 
 [features]
 testing = ["rand", "postcard", "serde_json", "serde-persistent-deserializer"]
--- a/manul/src/protocol/round.rs
+++ b/manul/src/protocol/round.rs
@@ -369,6 +369,25 @@ pub trait PartyId: 'static + Debug + Clone + Ord + Send + Sync + Serialize + for
 
 impl<T> PartyId for T where T: 'static + Debug + Clone + Ord + Send + Sync + Serialize + for<'de> Deserialize<'de> {}
 
+/// The specific way the node participates in the echo round (if any).
+#[derive(Debug, Clone)]
+pub enum EchoRoundParticipation<Id> {
+    /// The default behavior: sends broadcasts and receives echoed messages, or does neither.
+    ///
+    /// That is, this node will be a part of the echo round if [`Round::make_echo_broadcast`] generates a message.
+    Default,
+
+    /// This node sends broadcasts that will be echoed, but does not receive any.
+    Send,
+
+    /// This node receives broadcasts that it needs to echo, but does not send any itself.
+    Receive {
+        /// The other participants of the echo round
+        /// (that is, the nodes to which echoed messages will be sent).
+        echo_targets: BTreeSet<Id>,
+    },
+}
+
 /**
 A type representing a single round of a protocol.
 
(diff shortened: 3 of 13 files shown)


manul/src/protocol/round.rs
```diff
@@ -322,6 +322,25 @@ pub trait PartyId: 'static + Debug + Clone + Ord + Send + Sync {}
 
 impl<T> PartyId for T where T: 'static + Debug + Clone + Ord + Send + Sync {}
 
+/// The specific way the node participates in the echo round (if any).
+#[derive(Debug, Clone)]
+pub enum EchoRoundParticipation<Id> {
+    /// The default behavior: sends broadcasts and receives echoed messages, or does neither.
+    ///
+    /// That is, this node will be a part of the echo round if [`Round::make_echo_broadcast`] generates a message.
+    Default,
```
dvdplm, line 331: Would `Full` be a good name here? It doesn't fully capture the semantics but is at least a hint in the right direction whereas "default" doesn't mean anything unless the reader knows what the default is.

dvdplm (commented):
I'm struggling a bit to review this. The code seems all ok, modulo nitpicks, but I'm kind of stuck in my head on why we need this (despite the great PR description). E.g. in the `partial_echo` test, how should I think about a real world scenario there? Would node 0 be like a "dealer" of some sort, broadcasting but not participating? And what would node 4 be, given it's not participating at all?

clippy.toml
```diff
@@ -1 +0,0 @@
-allow-unwrap-in-tests = true
```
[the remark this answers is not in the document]
fjarri: For whatever reason, it is not.

manul/src/session/session.rs
```diff
@@ -97,6 +97,13 @@ impl AsRef<[u8]> for SessionId {
     }
 }
 
+#[derive(Debug)]
+pub(crate) struct EchoRoundInfo<Verifier> {
+    pub(crate) message_destinations: BTreeSet<Verifier>,
+    pub(crate) expecting_messages_from: BTreeSet<Verifier>,
+    pub(crate) expected_echos: BTreeSet<Verifier>,
```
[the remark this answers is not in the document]
fjarri: It wouldn't be a little ambiguous I think. `expected_echos` is the set of echos you're expecting to find in a pack that's sent to you (from any node), and `expecting_echos_from` could be understood as the nodes that need to send the echo packs to you. These can be different.

manul/src/protocol/round.rs
fjarri: Funny you should say that, `Full` was the original name I had here :) But I changed it to `Default` because it also covers the nodes which don't send or receive anything (node 4 in the test), where I wasn't sure `Full` was intuitive.

manul/src/session/session.rs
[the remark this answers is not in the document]
dvdplm: Hmm. Maybe just `echoes` works then?

manul/src/protocol/round.rs
dvdplm: `Full` isn't perfect, but gives a hint to the right understanding. I like it better than `Default`.

We could go with a `FullOrAbsent` but that reads very confusingly. :)

manul/Cargo.toml
```diff
@@ -20,6 +20,7 @@ rand_core = { version = "0.6.4", default-features = false }
 tracing = { version = "0.1", default-features = false }
 displaydoc = { version = "0.2", default-features = false }
 derive-where = "1"
+smallvec = { version = "1", default-features = false, features = ["serde"] }
```
dvdplm, line 23: There's an argument to prefer `tinyvec` to `smallvec`, given the former has a "no unsafe" policy (and the latter has had a few memory related bugs over the years).

Another tangential thought: RustCrypto use the [hybrid_array](https://crates.io/crates/hybrid-array) so we probably have it in the dependency tree. I haven't looked into it but perhaps it can be used here too?

manul/src/protocol/round.rs
```diff
@@ -29,24 +30,18 @@ pub enum FinalizeOutcome<Id: PartyId, P: Protocol> {
     Result(P::Result),
 }
 
-// Maximum depth of group nesting in RoundIds.
-// We need this to be limited to allow the nesting to be performed in `const` context
-// (since we cannot use heap there).
-const ROUND_ID_DEPTH: usize = 8;
-
 /// A round identifier.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
```
dvdplm, line 73: It's too bad SmallVec isn't `Copy`. :/

dvdplm (approved):

manul/Cargo.toml
fjarri: `hybrid_array` deals exclusively with stack arrays, I think.

PR #68, opened 2024-11-09
Stateful entry points

### LINKED ISSUE #64: Stateful entry points
The current round initializers are stateless. We have some object implementing a trait with an associated `Inputs` type, and a `new()` method that returns a `BoxedRound`. Which means we need to impl this trait for some other type, so now our entry point requires two types, and the latter type won't even be ever instantiated.

Instead, we could make the entry point stateful:
```rust
trait EntryPoint<Id> {
    type Protocol;
    fn make_round(self) -> BoxedRound<Id, Self::Protocol>;
}  
```

This would mean that for any protocol, the only exported entities would be the protocol type, and the entry point type (or could they even be the same type?). And the entry point type could have custom constructors defined, if some were required for convenience, or to perform pre-checks of inputs. 

One could argue that the first round could implement `EntryPoint` as well. But if it implements some public trait, all the associated types would have to be public too (specifically, if we add a typed `Round` trait). With a separate entry point, the rest of the implementation can be crate-private.

Note that some thought will be required to write new bounds for the chain combinator - we will have to be able to create an entry point for the second protocol from the first entry point and the first protocol's result.

### LINKED ISSUE #66: Determining that a round may produce a result
In the current API, a round defines a `fn possible_next_rounds(&self) -> BTreeSet<RoundId>` method, which returns an empty set if there are no possible next rounds. This is a sufficient condition to tell that this round produces a result, but not a necessary one - in some protocols we may either produce a result or enter an error round, in which case the returned set won't be empty.

This is important for the `chain` combinator from #60, because the empty `possible_next_rounds()` is how it determines when the first protocol ends. We need to know this to amend `possible_next_rounds()` to return the first round of the second protocol, so that caching messages in a `Session` works correctly. 

So, the problem is: how do we signal that the round produces a result, in the least intrusive way? Should we add another method to `Round`, `may_produce_result()`, with a blanket implementation (returning `self.possible_next_rounds().is_empty()`)? Or is there some way that will better guarantee that users don't forget to implement it if their protocol behaves differently?

fjarri:
Stacked on top of #67

This PR makes `EntryPoint`  types stateful - that is, this removes the need for `EntryPoint::Inputs` because the `EntryPoint`-implementing type becomes the inputs itself. Fixes #64.

Benefits:
- No need to expose `Round` types, only the `EntryPoint` impl can be public
- Protocol writers can write custom constructors (basically what we do in `synedrion` in `constructors.rs`) with additional consistency checks or convenience methods

Also added `Round::may_produce_result()`. Fixes #66

### DIFF
--- a/examples/src/simple_chain.rs
+++ b/examples/src/simple_chain.rs
@@ -1,37 +1,69 @@
+use alloc::collections::BTreeSet;
 use core::fmt::Debug;
 
 use manul::{
-    combinators::chain::{Chained, ChainedEntryPoint},
-    protocol::PartyId,
+    combinators::{
+        chain::{Chain, ChainedJoin, ChainedProtocol, ChainedSplit},
+        CombinatorEntryPoint,
+    },
+    protocol::{PartyId, Protocol},
 };
 
-use super::simple::{Inputs, Round1};
+use super::simple::{SimpleProtocol, SimpleProtocolEntryPoint};
 
-pub struct ChainedSimple;
+/// A protocol that runs the [`SimpleProtocol`] twice, in sequence.
+/// Illustrates the chain protocol combinator.
+#[derive(Debug)]
+pub struct DoubleSimpleProtocol;
+
+impl ChainedProtocol for DoubleSimpleProtocol {
+    type Protocol1 = SimpleProtocol;
+    type Protocol2 = SimpleProtocol;
+}
+
+pub struct DoubleSimpleEntryPoint<Id> {
+    all_ids: BTreeSet<Id>,
+}
+
+impl<Id: PartyId> DoubleSimpleEntryPoint<Id> {
+    pub fn new(all_ids: BTreeSet<Id>) -> Self {
+        Self { all_ids }
+    }
+}
+
+impl<Id> CombinatorEntryPoint for DoubleSimpleEntryPoint<Id> {
+    type Combinator = Chain;
+}
+
+impl<Id> ChainedSplit<Id> for DoubleSimpleEntryPoint<Id>
+where
+    Id: PartyId,
+{
+    type Protocol = DoubleSimpleProtocol;
+    type EntryPoint = SimpleProtocolEntryPoint<Id>;
+    fn make_entry_point1(self) -> (Self::EntryPoint, impl ChainedJoin<Id, Protocol = Self::Protocol>) {
+        (
+            SimpleProtocolEntryPoint::new(self.all_ids.clone()),
+            DoubleSimpleProtocolTransition { all_ids: self.all_ids },
+        )
+    }
+}
 
 #[derive(Debug)]
-pub struct NewInputs<Id>(Inputs<Id>);
+struct DoubleSimpleProtocolTransition<Id> {
+    all_ids: BTreeSet<Id>,
+}
 
-impl<'a, Id: PartyId> From<&'a NewInputs<Id>> for Inputs<Id> {
-    fn from(source: &'a NewInputs<Id>) -> Self {
-        source.0.clone()
+impl<Id> ChainedJoin<Id> for DoubleSimpleProtocolTransition<Id>
+where
+    Id: PartyId,
+{
+    type Protocol = DoubleSimpleProtocol;
+    type EntryPoint = SimpleProtocolEntryPoint<Id>;
+    fn make_entry_point2(self, _result: <SimpleProtocol as Protocol>::Result) -> Self::EntryPoint {
+        SimpleProtocolEntryPoint::new(self.all_ids)
     }
 }
-
-impl<Id: PartyId> From<(NewInputs<Id>, u8)> for Inputs<Id> {
-    fn from(source: (NewInputs<Id>, u8)) -> Self {
-        let (inputs, _result) = source;
-        inputs.0
-    }
-}
-
-impl<Id: PartyId> Chained<Id> for ChainedSimple {
-    type Inputs = NewInputs<Id>;
-    type EntryPoint1 = Round1<Id>;
-    type EntryPoint2 = Round1<Id>;
-}
-
-pub type DoubleSimpleEntryPoint<Id> = ChainedEntryPoint<Id, ChainedSimple>;
 
 #[cfg(test)]
 mod tests {
@@ -39,13 +71,12 @@ mod tests {
 
     use manul::{
         session::{signature::Keypair, SessionOutcome},
-        testing::{run_sync, BinaryFormat, TestSessionParams, TestSigner, TestVerifier},
+        testing::{run_sync, BinaryFormat, TestSessionParams, TestSigner},
     };
     use rand_core::OsRng;
     use tracing_subscriber::EnvFilter;
 
-    use super::{DoubleSimpleEntryPoint, NewInputs};
-    use crate::simple::Inputs;
+    use super::DoubleSimpleEntryPoint;
 
     #[test]
     fn round() {
@@ -54,24 +85,16 @@ mod tests {
             .iter()
             .map(|signer| signer.verifying_key())
             .collect::<BTreeSet<_>>();
-        let inputs = signers
+        let entry_points = signers
             .into_iter()
-            .map(|signer| {
-                (
-                    signer,
-                    NewInputs(Inputs {
-                        all_ids: all_ids.clone(),
-                    }),
-                )
-            })
+            .map(|signer| (signer, DoubleSimpleEntryPoint::new(all_ids.clone())))
             .collect::<Vec<_>>();
 
         let my_subscriber = tracing_subscriber::fmt()
             .with_env_filter(EnvFilter::from_default_env())
             .finish();
         let reports = tracing::subscriber::with_default(my_subscriber, || {
-            run_sync::<DoubleSimpleEntryPoint<TestVerifier>, TestSessionParams<BinaryFormat>>(&mut OsRng, inputs)
-                .unwrap()
+            run_sync::<_, TestSessionParams<BinaryFormat>>(&mut OsRng, entry_points).unwrap()
         });
 
         for (_id, report) in reports {
--- a/manul/src/combinators/chain.rs
+++ b/manul/src/combinators/chain.rs
@@ -4,42 +4,42 @@ executes the two inner protocols in sequence, feeding the result of the first pr
 into the inputs of the second protocol.
 
 For the session level users (that is, the ones executing the protocols)
-the new protocol is a single entity with its own [`Protocol`](`crate::protocol::Protocol`) type
-and an [`EntryPoint`](`crate::protocol::EntryPoint`) type.
+the new protocol is a single entity with its own [`Protocol`](`crate::protocol::Protocol`)-implementing type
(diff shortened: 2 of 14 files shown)


manul/src/combinators/chain.rs
```diff
@@ -4,42 +4,42 @@ executes the two inner protocols in sequence, feeding the result of the first pr
...
+- Internally, `EntryPointB` is created from `ResultA` and the data created in [`ChainedSplit::make_entry_point1`]
+  using the [`ChainedJoin`] implementation provided by the protocol author;
 - `RB1` is executed;
-- `RB2` is executed, producing `ResultB` (which is also the result of `ChainedProtocol`).
+- `RB2` is executed, producing `ResultB` (which is also the result of `ProtocolC`).
 
 If the execution happens in a [`Session`](`crate::session::Session`), and there is an error at any point,
 a regular evidence or correctness proof are created using the corresponding types from the new `ProtocolC`.
 
 The usage is as follows.
```
dvdplm, line 30:
```suggestion
Usage
```

manul/src/combinators/chain.rs
```diff
@@ -229,118 +225,138 @@ where
...
+        result: <<Self::Protocol as ChainedProtocol>::Protocol1 as Protocol>::Result,
+    ) -> Self::EntryPoint;
 }
 
-impl<Id, C> EntryPoint<Id> for ChainedEntryPoint<Id, C>
+impl<Id, T> EntryPoint<Id> for T
 where
     Id: PartyId,
-    C: Chained<Id>,
+    T: ChainedSplit<Id> + CombinatorEntryPoint<Type = Chain>,
```
dvdplm, line 302:
```suggestion
    T: ChainedSplit<Id> + CombinatorEntryPoint<Kind = Chain>,
```

I think it'd be good to avoid the word "type" when we can.

manul/src/combinators/chain.rs
```diff
@@ -359,11 +375,11 @@ where
                     .map(|round_id| round_id.group_under(1))
                     .collect::<BTreeSet<_>>();
 
-                // If there are no next rounds, this is the result round.
-                // This means that in the chain the next round will be the entry round of the second protocol.
-                if next_rounds.is_empty() {
-                    next_rounds.insert(C::EntryPoint2::entry_round().group_under(2));
+                if round.as_ref().may_produce_result() {
```
dvdplm, line 378:
```suggestion
                if round.as_ref().produces_result() {
```

I think `produces_result` is a better name.

manul/src/combinators/chain.rs
```diff
@@ -359,11 +375,11 @@ where
                     .map(|round_id| round_id.group_under(1))
```
dvdplm, line 375: (commenting here but the comment concerns line 368 `fn possible_next_rounds(&self) -> BTreeSet<RoundId> {`)

The name here suggests that we may or may not get a set of round IDs back: why not return `Option`? If no rounds were returned, the `BTreeSet` is empty?

manul/src/combinators/markers.rs
```diff
@@ -0,0 +1,9 @@
+/// A trait for combinator markers.
```
dvdplm, line 1:
```suggestion
/// A marker trait for protocol combinators.
```

manul/src/combinators/markers.rs
```diff
@@ -0,0 +1,9 @@
...
+pub trait CombinatorType {}
+
+/// Marks an entry point for a protocol combinator result, so that [`EntryPoint`](`crate::protocol::EntryPoint`)
+/// could be derived automatically for it.
```
dvdplm, line 5:
```suggestion
/// can be derived automatically for it.
```

manul/src/combinators/markers.rs
```diff
@@ -0,0 +1,9 @@
...
+pub trait CombinatorEntryPoint {
+    /// The type of the combinator.
+    type Type: CombinatorType;
```
dvdplm, line 8:
```suggestion
    type Kind: CombinatorKind;
```

manul/src/combinators/misbehave.rs
```diff
@@ -114,20 +104,30 @@ where
     }
 }
 
-/// The inputs for the misbehaving rounds.
-#[derive_where::derive_where(Debug; <M::EntryPoint as EntryPoint<Id>>::Inputs)]
-pub struct MisbehavingInputs<Id, B, M>
+/// The new entry point for the misbehaving rounds.
+///
+/// Use as an entry point to run the session, with your ID, behavior `B` and the misbehavior definition `M` set.
```
dvdplm, line 109:
```suggestion
/// Use as an entry point to run the session, with your ID, the behavior `B` and the misbehavior definition `M` set.
```

manul/src/protocol/round.rs
```diff
@@ -369,6 +366,25 @@ pub trait PartyId: 'static + Debug + Clone + Ord + Send + Sync + Serialize + for
 
 impl<T> PartyId for T where T: 'static + Debug + Clone + Ord + Send + Sync + Serialize + for<'de> Deserialize<'de> {}
 
+/// The specific way the node participates in the echo round (if any).
+#[derive(Debug, Clone)]
+pub enum EchoRoundParticipation<Id> {
+    /// The default behavior: sends broadcasts and receives echoed messages, or does neither.
+    ///
+    /// That is, this node will be a part of the echo round if [`Round::make_echo_broadcast`] generates a message.
+    Default,
```
dvdplm, line 375: You didn't like `Full` for this? :)

manul/src/session/session.rs
```diff
@@ -177,10 +184,36 @@ where
 
         let message_destinations = round.as_ref().message_destinations().clone();
 
-        let possible_next_rounds = if echo_broadcast.payload().is_none() {
-            round.as_ref().possible_next_rounds()
-        } else {
+        let echo_round_participation = round.as_ref().echo_round_participation();
+
+        let round_sends_echo_broadcast = !echo_broadcast.payload().is_none();
```
dvdplm, line 189: My brain stumbles a bit on double negations ("not is none"). Would using `is_some()` work here?

manul/src/testing/run_sync.rs
```diff
@@ -86,25 +86,25 @@ where
     Ok((state, messages))
 }
 
-/// Execute sessions for multiple nodes concurrently, given the the inputs
-/// for the first round `R` and the signer for each node.
+/// Execute sessions for multiple nodes concurrently,
+/// given the the vector of the signer and the entry point for each node.
```
dvdplm, line 90:
```suggestion
/// given a vector of the signer and the entry point as a tuple for each node.
```

manul/src/tests/partial_echo.rs
```diff
@@ -0,0 +1,227 @@
...
+        .with_env_filter(EnvFilter::from_default_env())
+        .finish();
+    let reports = tracing::subscriber::with_default(my_subscriber, || {
+        run_sync::<_, TestSessionParams<BinaryFormat>>(&mut OsRng, entry_points).unwrap()
+    });
+
+    for (_id, report) in reports {
+        assert!(matches!(report.outcome, SessionOutcome::Result(_)));
+    }
+}
```
dvdplm, line 227: It's nice to see how clean this test is. :)

dvdplm (approved):
LGTM.

Left nitpicks and suggestions, feel free to ignore and/or defer to separate PR(s).

examples/src/simple_chain.rs
```diff
@@ -1,51 +1,80 @@
...
-    type Inputs = NewInputs<Id>;
-    type EntryPoint1 = Round1<Id>;
-    type EntryPoint2 = Round1<Id>;
+#[derive(Debug)]
+struct DoubleTransition<Id> {
+    all_ids: BTreeSet<Id>,
 }
 
-pub type DoubleSimpleEntryPoint<Id> = ChainedEntryPoint<Id, ChainedSimple>;
+impl<Id> ChainedJoin<Id> for DoubleTransition<Id>
```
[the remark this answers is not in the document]
fjarri: Yeah I'm not particularly fond of the name either, but it had to be something. We can fix it later.

manul/src/combinators/chain.rs
fjarri: It won't be accurate though: a round may either produce a result, or transition to another (error) round. We are only interested if there's a possibility that this round may terminate the protocol, so that we know what messages to expect next in a chained protocol situation.

manul/Cargo.toml
```diff
@@ -29,12 +29,14 @@ serde_json = { version = "1", default-features = false, features = ["alloc"], op
...
 [dev-dependencies]
 impls = "1"
+rand_core = { version = "0.6.4", default-features = false, features = ["getrandom"] }
 rand = { version = "0.8", default-features = false }
 serde_asn1_der = "0.8"
 criterion = "0.5"
 serde-persistent-deserializer = "0.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
 serde_json = { version = "1", default-features = false, features = ["alloc"] }
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
```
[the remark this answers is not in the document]
fjarri: Looks interesting, we can avoid adding a tracing-enabled `run_sync()` in #71

manul/src/combinators/chain.rs
```diff
@@ -4,42 +4,42 @@ executes the two inner protocols in sequence, feeding the result of the first pr
...
 
 If the execution happens in a [`Session`](`crate::session::Session`), and there is an error at any point,
 a regular evidence or correctness proof are created using the corresponding types from the new `ProtocolC`.
 
 The usage is as follows.
 
-1. Define an input type for the new joined protocol.
-   Most likely it will be a union between inputs of the first and the second protocol.
+1. Implement [`ChainedProtocol`] for a type of your choice. Usually it will be a ZST.
+   You will have to specify the two protocol types you want to chain.
```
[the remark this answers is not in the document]
fjarri: I am not sure "must" is appropriate when something will be already enforced by the compiler... Maybe we need to get Johnny to proofread our docs.

manul/src/combinators/chain.rs
fjarri: I am not sure. Maybe it's just a heuristic to not use `Option` when there's an entirely appropriate default value that can be used in lieu of `None`. But in this case I wouldn't be opposed to an `Option`. In fact, the whole `Round` API needs a little review (returning `None` or some special value - in this method, or with messages, payloads, or artifacts; returning refs vs values and so on), but I was delaying that until #4 and #69.

manul/src/combinators/markers.rs
fjarri: Changed to `Combinator`

manul/src/protocol/round.rs
fjarri: I am not sold on it, but maybe it's not that important since it'll probably never get specified explicitly anyway, being the default. Let's discuss naming in a separate PR.

manul/src/session/session.rs
fjarri: It's a method of `EchoBroadcast`, not `Option`, so there's currently no `is_some()`. Perhaps there should be. (also I think this is a change from #68)

PR #71, opened 2024-11-15
Dev infrastructure

fjarri:
Stacked on top of #68

- Added `ExecutionResult` and `run_sync_with_tracing()`, to remove some boilerplate from tests
- Renamed `testing` module and feature to `dev` to avoid confusion with `test`


manul/src/testing/run_sync.rs
```diff
@@ -166,3 +167,21 @@ where
 
     Ok(outcomes)
 }
+
+/// Same as [`run_sync()`], but enables a [`tracing`] subscriber that prints the tracing events to stdout,
+/// taking options from the environment variable `RUST_LOG` (see [`mod@tracing_subscriber::fmt`] for details).
+#[allow(clippy::type_complexity)]
+pub fn run_sync_with_tracing<EP, SP>(
```
dvdplm, line 173: What is the point of actually providing two flavours of this? Like, why can't we have only `run_sync` (and make it use tracing)?

fjarri: The basic `run_sync` is used for benchmarks. With `test-log` I guess we don't need to provide a special function anymore.

PR #72, opened 2024-11-15
Remove `CorrectnessProof`

### LINKED ISSUE #2: Add a test for correctness proof generation and checking
Basically, what do we do when the session returns `SessionOutcome::StalledWithProof`? The test should include assembling such proofs from all the nodes and verifying them as if it was a third-party adjudicator.

In a realistic scenario, what has to be attached to these proofs? We probably need the same system as with provable errors, where it can provide a list of rounds to save messages from, but the messages would be from this node, not from another node.

fjarri:
Stacked on top of #71

Closes #2

- Remove `CorrectnessProof`.
- Since `FinalizeError` only contains `LocalError` now, it is removed as well and replaced with just `LocalError`.

I think it is safe to assume that any attributable error in the protocol would occur at `receive_message()` stage. At least for now, I would like to see an example of a protocol where it is not the case.


dvdplm (approved):
Good stuff.

PR #75, opened 2024-11-19
Adjust re-exports

fjarri:
- Remove `protocol::digest` re-export.
- `digest` and `signature` are now re-exported from the top level instead of `session`.

### DIFF
--- a/examples/src/simple.rs
+++ b/examples/src/simple.rs
@@ -405,7 +405,7 @@ mod tests {
 
     use manul::{
         dev::{run_sync, BinaryFormat, TestSessionParams, TestSigner},
-        session::signature::Keypair,
+        signature::Keypair,
     };
     use rand_core::OsRng;
     use test_log::test;
--- a/examples/src/simple.rs
+++ b/examples/src/simple.rs
@@ -5,7 +5,11 @@ use alloc::{
 };
 use core::fmt::Debug;
 
-use manul::protocol::*;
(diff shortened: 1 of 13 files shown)


examples/src/simple.rs
```diff
@@ -5,7 +5,11 @@ use alloc::{
 };
 use core::fmt::Debug;
 
-use manul::protocol::*;
+use manul::protocol::{
+    Artifact, BoxedRound, Deserializer, DirectMessage, EchoBroadcast, EntryPoint, FinalizeOutcome, LocalError,
+    MessageValidationError, NormalBroadcast, PartyId, Payload, Protocol, ProtocolError, ProtocolMessagePart,
+    ProtocolValidationError, ReceiveError, Round, RoundId, Serializer,
+};
```
dvdplm, line 12: What do you think about having a `prelude` when the code stabilizes a bit that users can *-import, giving them most of the basic types?

dvdplm (approved):
lgtm

examples/src/simple.rs
fjarri: Basically `protocol` and `session` are such preludes already, for two different types of users.

fjarri: There is a slight problem if you try to glob-import both, because they share `LocalError` and `RemoteError`. But a) perhaps they shouldn't (see #25), and b) I can't imagine a case where it would be necessary.

PR #87, opened 2025-01-21
Allow overriding `finalize()` in `Misbehaving`

fjarri:
Stacked on top of #86

This came up in https://github.com/entropyxyz/synedrion/pull/170. In order to test the optional error round in InteractiveSigning, the malicious node needs to be forced to transition into it. But since it only receives valid messages from lawful nodes, it has no reason to. The new `Misbehaving::override_finalize()` allows one to replace the existing `Round::finalize()` with custom logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Exposed `dev::ExecutionResult`. ([#79])
 - `NoProtocolErrors` stub type to indicate that the protocol does not generate any provable errors. ([#79])
 - Conversion from `u8` to `RoundId` and comparison of `RoundId` with `u8`. ([#84])
+- `Misbehaving::override_finalize()` for malicious finalization logic. ([#87])
 
 
 ### Fixed
@@ -48,6 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#84]: https://github.com/entropyxyz/manul/pull/84
 [#85]: https://github.com/entropyxyz/manul/pull/85
 [#86]: https://github.com/entropyxyz/manul/pull/86
+[#87]: https://github.com/entropyxyz/manul/pull/87
 
 
 ## [0.1.0] - 2024-11-19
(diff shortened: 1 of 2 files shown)


manul/src/dev/run_sync.rs
```diff
@@ -27,6 +27,73 @@ struct RoundMessage<SP: SessionParameters> {
...
+            Self::Ordered(m) => m.entry(message.from.clone()).or_insert(Vec::new()).push(message),
+            Self::Unordered(v) => v.push(message),
+        }
+    }
+
+    /// Adds a a vector of messages to the queue.
+    fn extend(&mut self, messages: Vec<RoundMessage<SP>>) {
+        for message in messages {
+            self.push(message)
+        }
```
dvdplm, line 61: I guess you could use `Vec::extend` for the `Unordered` case, but I don't think this code is performance sensitive so I wouldn't bother.

manul/src/dev/run_sync.rs
```diff
@@ -27,6 +27,73 @@ struct RoundMessage<SP: SessionParameters> {
...
+    }
+
+    /// Removes a random message from the queue and returns it.
+    fn pop(&mut self, rng: &mut impl CryptoRngCore) -> RoundMessage<SP> {
+        match self {
+            Self::Ordered(m) => {
+                let senders = m.keys().cloned().collect::<Vec<_>>();
+                let sender_idx = rng.gen_range(0..senders.len());
+                let sender = &senders.get(sender_idx).expect("the entry exists");
```
dvdplm, line 70: Can you find a way to avoid cloning all the keys? Maybe something like `let sender = m.keys().nth(sender_idx)`?

dvdplm (approved):
LGTM modulo nitpicks that you can disregard if you want.

manul/src/dev/run_sync.rs
fjarri: Good idea, this works, I will modify #86

manul/src/dev/run_sync.rs
fjarri: Yeah, it's a minor performance improvement (since Vec won't reallocate on each push) versus minor readability improvement (reducing `extend()` to `push()`)

PR #90, opened 2025-02-01
Send hashes in the echo round instead of the full messages

fjarri:
Some echo broadcasts in CGGMP'24 include large structures that scale linearly with the number of nodes, so the corresponding echo round messages would scale quadratically. This PR replaces sending full payloads the second time with just sending payload hashes. The trade-offs are:
- every message needs two hash applications instead of one (hash the payload, then hash (payload hash, metadata))
- more information to keep in the `Evidence` (need to keep the echo messages themselves and the signed hashes)

Also fixes a bug with the `TestHasher` `OutputSize` and `buffer.len()` mismatch. Before this PR the created digest was actually ignored when creating `TestSignature`, so the bug didn't fire.

This approach hashes in two stages when it's technically not necessary (that is, for non-echo messages), but I don't think it leads to any noticeable performance decrease — with most hashes it's the update part that's slow, not the finalize. So the only additional cost is an extra tag pushed into the digest.

### DIFF
--- a/manul/src/dev/session_parameters.rs
+++ b/manul/src/dev/session_parameters.rs
@@ -62,7 +62,7 @@ impl<D: digest::Digest> signature::DigestVerifier<D, TestSignature> for TestVeri
 #[derive(Debug, Clone, Copy, Default)]
 pub struct TestHasher {
     cursor: usize,
-    buffer: [u8; 32],
+    buffer: digest::Output<Self>,
 }
 
 impl digest::HashMarker for TestHasher {}
@@ -72,7 +72,7 @@ impl digest::Update for TestHasher {
         // A very simple algorithm for testing, just xor the data in buffer-sized chunks.
         for byte in data {
             *self.buffer.get_mut(self.cursor).expect("index within bounds") ^= byte;
-            self.cursor = (self.cursor + 1) % 32;
+            self.cursor = (self.cursor + 1) % self.buffer.len();
         }
     }
 }
@@ -84,7 +84,7 @@ impl digest::FixedOutput for TestHasher {
 }
 
 impl digest::OutputSizeUser for TestHasher {
-    type OutputSize = typenum::U8;
+    type OutputSize = typenum::U32;
 }
 
 /// An implementation of [`SessionParameters`] using the testing signer/verifier types.
--- a/manul/src/session/message.rs
+++ b/manul/src/session/message.rs
@@ -11,12 +11,10 @@ use super::{
     wire_format::WireFormat,
     LocalError,
 };
-use crate::protocol::{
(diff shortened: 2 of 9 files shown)


manul/src/session/message.rs
```diff
@@ -78,19 +78,31 @@ pub struct MessageWithMetadata<M> {
     message: M,
 }
 
+fn payload_hash<SP: SessionParameters>(message: &impl ProtocolMessagePart) -> digest::Output<SP::Digest> {
+    let digest = SP::Digest::new_with_prefix(b"PayloadHash");
+    let digest = match message.maybe_message().as_ref() {
+        None => digest.chain_update([0u8]),
```
dvdplm, line 84: This mean that we return the same hash for all empty messages, regardless of the message (`DirectMessage`, `EchoBroadcast` and `NormalBroadcast`) – is that ok or could an attacker substitute one message for the other?

manul/src/session/message.rs
```diff
@@ -324,3 +348,60 @@ impl<Verifier> VerifiedMessage<Verifier> {
...
+            .map_err(|_| MessageVerificationError::InvalidSignature)?;
+        if verifier.verify_digest(digest, &signature).is_ok() {
+            Ok(VerifiedMessageHash {
+                signature: self.signature,
+                metadata: self.metadata,
+                payload_hash: self.payload_hash,
+            })
+        } else {
+            Err(MessageVerificationError::SignatureMismatch)
+        }
```
dvdplm, line 383:
```suggestion
        verifier
            .verify_digest(digest, &signature)
            .map(|r| VerifiedMessageHash {
                signature: self.signature,
                metadata: self.metadata,
                payload_hash: self.payload_hash,
            })
            .map_err(|_e| MessageVerificationError::SignatureMismatch)
```

If we had a `From<signature::Error> for MessageVerificationError` impl the last line could go away too.

dvdplm (approved):
LGTM

fjarri:
> This optimization is CGGMP-specific

I think it's a pretty general thing, we don't want to send whole messages the second time. Even without quadratic scaling echos with full messages can get pretty large.

manul/src/dev/session_parameters.rs
```diff
@@ -84,7 +84,7 @@ impl digest::FixedOutput for TestHasher {
 }
 
 impl digest::OutputSizeUser for TestHasher {
-    type OutputSize = typenum::U8;
+    type OutputSize = typenum::U32;
```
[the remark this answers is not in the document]
fjarri: Changed `buffer` type to `digest::Output<Self>` to avoid size duplication.

manul/src/session/message.rs
fjarri: Good point, I think they should be different.

Issue #13, opened 2024-10-12
Relax `Send`/`Sync` requirements


fjarri:
If someone doesn't offload message processing into tasks, the types don't need to be `Send`/`Sync`. Is it possible to gate it under a feature somehow? Is it actually a useful feature or an extremely narrow use case?

PR #91, opened 2025-02-06
Add a `session::tokio` module with a convenience function for executing a session

### LINKED ISSUE #22: Extend `example/tests/async.rs` to offload message creation/processing to separate tasks
This will demonstrate the target usage pattern for a heavy protocol.

Also I have a suspicion that we won't actually be able to `finalize()` because it consumes the `Session` object, and references to it are being sent to tasks, so we will have to wait for all of them to finish, which we won't want to do if there's a threshold needed to finalize (see also #11). So we will need to think of some different API to support this case.

### LINKED ISSUE #28: A session loop implementation for `tokio`
It is not trivial to write a session loop, especially with offloading of creating/verifying messages to tasks. We can supply at least a `tokio`-based implementation, gated behind a feature. `example/tests/async.rs` can be used as a starting point.

fjarri:
- Added `session::tokio::run_session()` and supporting types. Gated behind the `tokio` feature. Fixes #28
- Added `session::tokio::par_run_session()` that runs the session with offloading message processing into spawned tasks. Fixes #22 
- Added a benchmark for message offloading
- Moved the async running machinery from `examples/tests/async_runner` into `manul::dev::tokio`

Regarding the info included the channels and included info, for now we will:
- take channels by reference;
- include session ID in the outgoing message (since it can be a sink given to multiple sessions), but not have one in `MessageIn` (since that end of the channel is necessarily personalized)

We will get `entropy-core` to use it and iterate on the results.

There are some weird `Send`/`Sync` bounds, I will attempt to normalize them as a part of #13.

### DIFF
--- a/manul/src/session/tokio.rs
+++ b/manul/src/session/tokio.rs
@@ -1,0 +1,174 @@
+//! High-level API for executing sessions in `tokio` tasks.
+
+use alloc::{format, vec::Vec};
+
+use rand_core::CryptoRngCore;
+use tokio::sync::mpsc;
+use tracing::{debug, trace};
+
+use super::{
+    message::Message,
+    session::{CanFinalize, RoundOutcome, Session, SessionId, SessionParameters},
+    transcript::SessionReport,
+    LocalError,
+};
+use crate::protocol::Protocol;
+
+/// The outgoing message from a local session.
+#[derive(Debug)]
+pub struct MessageOut<SP: SessionParameters> {
+    /// The session ID that created the message.
+    ///
+    /// Useful when there are several sessions running on a node, pushing messages into the same channel.
+    pub session_id: SessionId,
+    /// The verifying key of the party that created the message.
+    ///
+    /// Useful when there are several sessions running on a node, pushing messages into the same channel.
+    pub from: SP::Verifier,
+    /// The verifying key of the party the message is intended for.
+    pub to: SP::Verifier,
+    /// The message to be sent.
+    ///
+    /// Note that the caller is responsible for encrypting the message and attaching authentication info.
+    pub message: Message<SP::Verifier>,
+}
+
+/// The incoming message from a remote session.
+#[derive(Debug)]
+pub struct MessageIn<SP: SessionParameters> {
+    /// The verifying key of the party the message originated from.
+    ///
+    /// It is assumed that the message's authentication info has been checked at this point.
+    pub from: SP::Verifier,
+    /// The incoming message.
+    pub message: Message<SP::Verifier>,
+}
+
+/// Executes the session waiting for the messages from the `rx` channel
+/// and pushing outgoing messages into the `tx` channel.
+pub async fn run_session<P, SP>(
+    rng: &mut impl CryptoRngCore,
+    tx: &mpsc::Sender<MessageOut<SP>>,
+    rx: &mut mpsc::Receiver<MessageIn<SP>>,
+    session: Session<P, SP>,
+) -> Result<SessionReport<P, SP>, LocalError>
+where
+    P: Protocol<SP::Verifier>,
+    SP: SessionParameters,
+{
+    let mut session = session;
+    // Some rounds can finalize early and put off sending messages to the next round. Such messages
+    // will be stored here and applied after the messages for this round are sent.
+    let mut cached_messages = Vec::new();
+
+    let my_id = format!("{:?}", session.verifier());
+
+    // Each iteration of the loop progresses the session as follows:
+    //  - Send out messages as dictated by the session "destinations".
+    //  - Apply any cached messages.
+    //  - Enter a nested loop:
+    //      - Try to finalize the session; if we're done, exit the inner loop.
+    //      - Wait until we get an incoming message.
+    //      - Process the message we received and continue the loop.
+    //  - When all messages have been sent and received as specified by the protocol, finalize the
+    //    round.
+    //  - If the protocol outcome is a new round, go to the top of the loop and start over with a
+    //    new session.
+    loop {
+        debug!("{my_id}: *** starting round {:?} ***", session.round_id());
+
+        // This is kept in the main task since it's mutable,
+        // and we don't want to bother with synchronization.
+        let mut accum = session.make_accumulator();
+
+        // Note: generating/sending messages and verifying newly received messages
+        // can be done in parallel, with the results being assembled into `accum`
+        // sequentially in the host task.
+
+        let destinations = session.message_destinations();
+        for destination in destinations.iter() {
+            // In production usage, this will happen in a spawned task
+            // (since it can take some time to create a message),
+            // and the artifact will be sent back to the host task
+            // to be added to the accumulator.
+            let (message, artifact) = session.make_message(rng, destination)?;
+            debug!("{my_id}: Sending a message to {destination:?}",);
+            tx.send(MessageOut {
+                session_id: session.session_id().clone(),
+                from: session.verifier().clone(),
+                to: destination.clone(),
+                message,
+            })
+            .await
+            .map_err(|err| {
+                LocalError::new(format!(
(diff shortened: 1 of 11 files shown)


fjarri:
@ameba23 this is supposed to be the replacement for https://github.com/entropyxyz/entropy-core/blob/master/crates/protocol/src/execute_protocol.rs#L73 so that you don't have to worry about executing the session correctly. Does the API look okay to you? I noticed that you're taking the channels by value and then returning them; will the proposed API with references work?

manul/src/session/tokio.rs
```diff
@@ -0,0 +1,169 @@
...
+    message::Message,
+    session::{CanFinalize, RoundOutcome, Session, SessionParameters},
+    transcript::SessionReport,
+    LocalError,
+};
+use crate::protocol::Protocol;
+
+/// The outgoing message from a local session.
+#[derive(Debug)]
+pub struct MessageOut<SP: SessionParameters> {
```
dvdplm, line 19: The `MessageOut` and `MessageIn` type do not seem to be tokio-specific. Are they generally useful you reckon or will users want to implement their own versions? If the latter, why would they want that?

manul/src/session/tokio.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// The ID of the session that created the message.
```
dvdplm, line 20: Is this accurate? Given there is a public `SessionId` type, it is a bit confusing to read this and then read `SP::Verifier` which boils down to a `PartyId`. Is it "The ID of the party that created the message in this session" perhaps?

manul/src/session/tokio.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// The message to be sent.
+    ///
+    /// Note that the caller is responsible for encrypting the message and attaching authentication info.
+    pub message: Message<SP::Verifier>,
+}
+
+/// The incoming message from a remote session.
+#[derive(Debug)]
+pub struct MessageIn<SP: SessionParameters> {
+    /// The ID of the session the message originated from.
```
dvdplm, line 35: Same as above

manul/src/session/tokio.rs
```diff
@@ -0,0 +1,169 @@
...
+where
+    P: Protocol<SP::Verifier>,
+    SP: SessionParameters,
+{
+    let mut session = session;
+    // Some rounds can finalize early and put off sending messages to the next round. Such messages
+    // will be stored here and applied after the messages for this round are sent.
+    let mut cached_messages = Vec::new();
+
+    let key = session.verifier();
```
dvdplm, line 60:
```suggestion
    let verifier = session.verifier();
```
…or something more specific to its role? Maybe just `from`?

manul/src/session/tokio.rs
```diff
@@ -0,0 +1,169 @@
...
+            tx.send(MessageOut {
+                from: session.verifier().clone(),
+                to: destination.clone(),
+                message,
+            })
+            .await
+            .map_err(|err| {
+                LocalError::new(format!(
+                    "Failed to send a message from {:?} to {:?}: {err}",
+                    session.verifier(),
```
dvdplm, line 101:
```suggestion
                    key,
```

dvdplm (approved):
lgtm modulo questions and nitpicks

fjarri:
> the payloads of these channels currently contain both the synedrion message and the session ID

This can be included in `MessageOut`, but it would be strange in `MessageIn`. Do you need it to be present in both directions?

> The only other thing that we would lose by using this is that we are calling session.make_message and session.process_message in a separate tokio task. 

That's exactly #22 which I mentioned above, so it'll be included as well. I saw that you have some offloading in place, but I wonder, wouldn't it be better to use `spawn_blocking` there?

ameba23:
> This can be included in `MessageOut`, but it would be strange in `MessageIn`. Do you need it to be present in both directions?

It seems like we do need it in `MessageIn`. Sometimes a remote party sends the first message on the next protocol before we have finalized the current one. I'll double check but im pretty sure if i take that check out the tests will fail.
 
> That's exactly #22 which I mentioned above, so it'll be included as well. 

Thats great!

> I saw that you have some offloading in place, but I wonder, wouldn't it be better to use `spawn_blocking` there?

I've never used `spawn_blocking` before, but i guess i'll try it and see if it improves things.

fjarri:
> So if synedrion were to offer these high level protocols in a single session, we would not need to do this and from our point of view everything would be much simpler.

That may be actually possible using the chain combinator we have. I will look into it after this PR. 

> It seems like we do need it in MessageIn. Sometimes a remote party sends the first message on the next protocol before we have finalized the current one. I'll double check but im pretty sure if i take that check out the tests will fail.

But still, this situation should be handled a level above, shouldn't it? If you are routing the messages to several sessions, the session loop should not be given a message with a session ID that is known to not belong to that session.

> I've never used spawn_blocking before, but i guess i'll try it and see if it improves things.

So I wrote an implementation of an offloading executor (pushed to this PR), and according to my preliminary benchmarks it is faster than the non-offloading version. Benchmarks are very slow for some reason, trying to figure out why, so the exact numbers are still pending.

fjarri:
Profiling update: 10 nodes, 5 rounds, making a message and verifying a message does some CPU work (modular multiplication on repeat). No offloading: 14s, with offloading: 1.5s. I think that's pretty good. 

Note that, of course, it's a little tricky to benchmark. I'm running multiple nodes in a single-threaded tokio runtime, so in both cases the main execution thread of each node is still executed one after the other, which will be different if every node is running on a separate machine (although it would be interesting to test that too). But I think the results still indicate that offloading is pretty useful for computation-heavy protocols.

fjarri:
@ameba23, I am wondering if passing the channels through and attaching session IDs can be done externally? That is, instead of spawning `run_session()`, you would spawn something like 
```rust
async move {
    let mut rx = rx;
    let tx = tx;
    let result = run_session(&mut rx, &tx, ...)?;
    Ok((rx, tx, result))
}
```
which would take care of the channels.

It's trickier with attaching the session ID. I was hoping there's something in `tokio` that would allow you to attach a "mapping" to a `Sender` or `Receiver` before giving it to `run_session` that would translate between `MessageIn`/`MessageOut` and whatever extended format you need. But it doesn't seem that way.

ameba23:
> But still, this situation should be handled a level above, shouldn't it? If you are routing the messages to several sessions, the session loop should not be given a message with a session ID that is known to not belong to that session.

Yes, probably thats what makes the most sense.  I think it means we'll have to have separate channels for each sub-session going to a single websocket connection handler. But that is doable.

> Profiling update: 10 nodes, 5 rounds, making a message and verifying a message does some CPU work (modular multiplication on repeat). No offloading: 14s, with offloading: 1.5s. I think that's pretty good.

Wow thats great, i really hope this is what we will see in production.  I also tried doing this here: https://github.com/entropyxyz/entropy-core/pull/1285 but it did not appear to give a performance improvement using the signing protocol and 8 parties. But i am also not sure the way i am benchmarking is accurate. Either way, i agree that this is better - we should not be using `tokio::spawn` for synchronous tasks which are computationally heavy.

ameba23 (approved):

fjarri:
> I also tried doing this here: https://github.com/entropyxyz/entropy-core/pull/1285 but it did not appear to give a performance improvement using the signing protocol and 8 parties.

if you're not running it in a single-threaded Tokio runtime, all the CPU cores are already used by sessions on each node, so parallelizing doesn't do much.

manul/src/session/tokio.rs
fjarri: Renamed to `my_id`, and pre-formatted.

manul/src/session/tokio.rs
fjarri: We can generalize them later if we add other async runtime support

PR #93, opened 2025-02-11
Merge `Round` methods

### LINKED ISSUE #69: Merge some of the `Round` methods
`Round` has several methods that deal with the round's position in the state transition graph:
- `id(&self) -> RoundId`,
- `fn possible_next_rounds(&self) -> BTreeSet<RoundId>`,
- with #66 fixed, possibly some `fn returns_result(&self) -> bool`.
- to handle #78, we may need `fn possible_parallel_rounds(&self) -> BTreeSet<RoundId>`

Similarly, there are three methods dealing with how the node interacts with other nodes during the round (intersects with #4):
- `fn message_destinations(&self) -> &BTreeSet<Id>`,
- `fn expecting_messages_from(&self) -> &BTreeSet<Id>`,
- to be added in #67, `fn echo_round_participation(&self) -> EchoRoundParticipation<Id>`.

Does it make sense to merge either of the group into one method returning an enum? Which possibilities do we need to cover?

### LINKED ISSUE #78: More consistent handling of out-of-order messages
Suppose a protocol has a round A that can transition either to round B or round C. If Node 1 is in round A, and receives a message for round B (from another node who's already in it), but later transitions to round C, the message is discarded with no repercussions. But if Node 1 is already in round C, the message is treated as an unprovable malicious action and the event is recorded.

For consistency, these situations should be treated the same. Ideally, both events should not be recorded. To do that, when transitioning to round C we need to somehow pass on the information that another possible round other nodes can be in is round B, so we shouldn't treat messages from it as malicious actions.

fjarri:
- Merge `Round::id()`, `possible_next_rounds()` and `may_produce_result()` into `transition_info()`. Fixes #78
- Merge `Round::message_destinations()`, `expecting_messages_from()` and `echo_round_participation()` into `communication_info()`. With the previous item, fixes #69.

### DIFF
--- a/manul/src/protocol/round.rs
+++ b/manul/src/protocol/round.rs
@@ -5,17 +5,17 @@ use alloc::{
 };
 use core::{
     any::Any,
-    fmt::{self, Debug, Display},
+    fmt::{Debug, Display},
 };
 
 use rand_core::CryptoRngCore;
 use serde::{Deserialize, Serialize};
-use tinyvec::TinyVec;
 
 use super::{
     errors::{LocalError, MessageValidationError, ProtocolValidationError, ReceiveError},
     message::{DirectMessage, EchoBroadcast, NormalBroadcast, ProtocolMessage, ProtocolMessagePart},
     object_safe::BoxedRound,
+    round_id::{RoundId, TransitionInfo},
     serialization::{Deserializer, Serializer},
 };
 
@@ -26,132 +26,6 @@ pub enum FinalizeOutcome<Id: PartyId, P: Protocol<Id>> {
     AnotherRound(BoxedRound<Id, P>),
     /// The protocol reached a result.
     Result(P::Result),
-}
-
-/// A round identifier.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
-pub struct RoundId {
-    round_nums: TinyVec<[u8; 4]>,
-    is_echo: bool,
-}
-
-impl Display for RoundId {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        write!(f, "Round ")?;
-        for (i, round_num) in self.round_nums.iter().enumerate().rev() {
-            write!(f, "{}", round_num)?;
-            if i != 0 {
-                write!(f, "-")?;
-            }
-        }
-        if self.is_echo {
-            write!(f, " (echo)")?;
-        }
-        Ok(())
-    }
-}
-
-impl RoundId {
-    /// Creates a new round identifier.
-    pub fn new(round_num: u8) -> Self {
-        let mut round_nums = TinyVec::new();
-        round_nums.push(round_num);
-        Self {
-            round_nums,
-            is_echo: false,
-        }
-    }
-
-    /// Prefixes this round ID (possibly already nested) with a group number.
-    pub(crate) fn group_under(&self, round_num: u8) -> Self {
-        let mut round_nums = self.round_nums.clone();
-        round_nums.push(round_num);
-        Self {
-            round_nums,
-            is_echo: self.is_echo,
-        }
-    }
-
-    /// Removes the top group prefix from this round ID
-    /// and returns this prefix along with the resulting round ID.
-    ///
-    /// Returns the `Err` variant if the round ID is not nested.
-    pub(crate) fn split_group(&self) -> Result<(u8, Self), LocalError> {
-        if self.round_nums.len() == 1 {
-            Err(LocalError::new("This round ID is not in a group"))
-        } else {
-            let mut round_nums = self.round_nums.clone();
-            let group = round_nums.pop().expect("vector size greater than 1");
-            let round_id = Self {
-                round_nums,
-                is_echo: self.is_echo,
-            };
-            Ok((group, round_id))
-        }
-    }
-
-    /// Removes the top group prefix from this round ID and returns the resulting Round ID.
-    ///
-    /// Returns the `Err` variant if the round ID is not nested.
-    pub(crate) fn ungroup(&self) -> Result<Self, LocalError> {
-        if self.round_nums.len() == 1 {
-            Err(LocalError::new("This round ID is not in a group"))
-        } else {
-            let mut round_nums = self.round_nums.clone();
-            round_nums.pop().expect("vector size greater than 1");
-            Ok(Self {
-                round_nums,
-                is_echo: self.is_echo,
-            })
-        }
-    }
-
-    /// Returns `true` if this is an ID of an echo broadcast round.
-    pub(crate) fn is_echo(&self) -> bool {
-        self.is_echo
-    }
-
-    /// Returns the identifier of the echo round corresponding to the given non-echo round.
-    ///
-    /// Panics if `self` is already an echo round identifier.
-    pub(crate) fn echo(&self) -> Self {
-        // If this panic happens, there is something wrong with the internal logic
-        // of managing echo-broadcast rounds.
-        if self.is_echo {
-            panic!("This is already an echo round ID");
-        }
-        Self {
-            round_nums: self.round_nums.clone(),
-            is_echo: true,
-        }
-    }
-
-    /// Returns the identifier of the non-echo round corresponding to the given echo round.
-    ///
-    /// Panics if `self` is already a non-echo round identifier.
-    pub(crate) fn non_echo(&self) -> Self {
-        // If this panic happens, there is something wrong with the internal logic
-        // of managing echo-broadcast rounds.
-        if !self.is_echo {
-            panic!("This is already an non-echo round ID");
-        }
-        Self {
-            round_nums: self.round_nums.clone(),
-            is_echo: false,
-        }
-    }
-}
-
-impl From<u8> for RoundId {
-    fn from(source: u8) -> Self {
-        Self::new(source)
-    }
-}
-
-impl PartialEq<u8> for RoundId {
-    fn eq(&self, rhs: &u8) -> bool {
-        self == &RoundId::new(*rhs)
-    }
 }
 
 /// A distributed protocol.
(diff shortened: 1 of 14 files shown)


manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub fn new(round_num: u8) -> Self {
+        let mut round_nums = TinyVec::new();
+        round_nums.push(round_num);
+        Self {
+            round_nums,
+            is_echo: false,
+        }
+    }
+
+    /// Prefixes this round ID (possibly already nested) with a group number.
```
dvdplm, line 43: What do you think about adding module docs explaining what a "group" is in this context?

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+            Err(LocalError::new("This round ID is not in a group"))
+        } else {
+            let mut round_nums = self.round_nums.clone();
+            round_nums.pop().expect("vector size greater than 1");
+            Ok(Self {
+                round_nums,
+                is_echo: self.is_echo,
+            })
+        }
+    }
```
dvdplm, line 85: This method is *very* similar to `split_group` above. Are you sure we need both?

If we do keep it, what do you think about making it take `&mut self` instead and mutate in-place? "ungroup" sounds like a destructive operation to me.

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    /// Returns `true` if this is an ID of an echo broadcast round.
+    pub(crate) fn is_echo(&self) -> bool {
+        self.is_echo
+    }
+
+    /// Returns the identifier of the echo round corresponding to the given non-echo round.
+    ///
+    /// Panics if `self` is already an echo round identifier.
```
dvdplm, line 94: Why is crashing the right thing to do here?

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+        self == &RoundId::new(*rhs)
+    }
+}
+
+/// Information about the position of the round in the state transition graph.
+#[derive(Debug, Clone)]
+pub struct TransitionInfo {
+    /// The round ID.
+    ///
+    /// **Note:** these should not repeat during execution.
```
dvdplm, line 140: So no loops? Perhaps worth mentioning in the top-level docs?

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// For example, a round can be followed either by a happy path round or an error round,
+    /// depending on the outcome of some checks during the finalization.
+    pub siblings: BTreeSet<RoundId>,
+
+    /// The round IDs of the rounds this round can finalize into.
+    ///
+    /// Returns an empty set if this round only finalizes into a result.
+    pub children: BTreeSet<RoundId>,
+
+    /// ``true`` if this round's [`Round::finalize`](`crate::protocol::Round::finalize`)
```
dvdplm, line 157: DQ: why the double backticks?

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// may return [`FinalizeOutcome::Result`](`crate::protocol::FinalizeOutcome::Result`).
```
dvdplm, line 158: Is this correct?
- if false, this round will definitely not return a `FinalizeOutcome::Result`
- if true, this round may or may not return a `FinalizeOutcome::Result`

Is it correct to say that it *will* return a `FinalizeOutcome::Result` if the protocol was executed correctly, i.e. if there were no errors and this boolean is true, then this round will produce a result?

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+            siblings: self.siblings.into_iter().map(|r| r.group_under(group)).collect(),
+            children: self.children.into_iter().map(|r| r.group_under(group)).collect(),
+            may_produce_result: self.may_produce_result,
+        }
+    }
+
+    /// Returns the set of round IDs that can be simultaneously active on other nodes
+    /// (not including the current round ID).
+    ///
+    /// This includes: the children rounds (if some nodes already finalized this round),
```
dvdplm, line 177:
```suggestion
    /// This includes: the child rounds (if some nodes already finalized this round),
```

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// the parent rounds (if those nodes still haven't finalized while we already have),
```
dvdplm, line 178:
```suggestion
    /// the parent rounds (if those nodes still are not finalized while we already are),
```

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+            children: self.children,
+            may_produce_result: self.may_produce_result,
+        }
+    }
+
+    /// Creates a transition info for a non-terminating round in a linear sequence of rounds starting with 1.
+    ///
+    /// That is, if there are rounds 1, 2, 3, ..., N, where the N-th one returns the result,
+    /// this constructor can be used for rounds 1 to N-1.
+    pub fn regular(round_num: u8) -> Self {
```
dvdplm, line 210:
```suggestion
    pub fn new_regular(round_num: u8) -> Self {
```
Also: how about naming it `new_non_terminating` or `new_intermediate`?

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+            children: [RoundId::new(round_num + 1)].into(),
+            may_produce_result: false,
+        }
+    }
+
+    /// Creates a transition info for the result round in a linear sequence of rounds starting with 1.
+    ///
+    /// That is, if there are rounds 1, 2, 3, ..., N, where the N-th one returns the result,
+    /// this constructor can be used for round N.
+    pub fn result(round_num: u8) -> Self {
```
dvdplm, line 228: I think constructors should have the `new_` prefix (or `with_` or similar idiomatic constructor naming).

manul/src/protocol/round_id.rs
```diff
@@ -0,0 +1,267 @@
...
+
+impl PartialEq<u8> for RoundId {
+    fn eq(&self, rhs: &u8) -> bool {
+        self == &RoundId::new(*rhs)
+    }
+}
+
+/// Information about the position of the round in the state transition graph.
+#[derive(Debug, Clone)]
+pub struct TransitionInfo {
```
dvdplm, line 137: Alternative name suggestions: 
- `RoundState`: emphasizes that this struct describes the state of a round in the context of transitions.
- `RoundTopology`: emphasizes the interactions of this round with other rounds.

manul/src/protocol/round.rs
```diff
@@ -5,153 +5,64 @@ use alloc::{
...
-}
-
-impl PartialEq<u8> for RoundId {
-    fn eq(&self, rhs: &u8) -> bool {
-        self == &RoundId::new(*rhs)
-    }
+/// Possible successful outcomes of [`Round::finalize`].
+#[derive(Debug)]
+pub enum FinalizeOutcome<Id: PartyId, P: Protocol<Id>> {
+    /// Transition to a new round.
```
dvdplm, line 62:
```suggestion
    /// Transition to a new set of rounds. This is the case for protocols built using combinators, where a sub-protocol is a part of a bigger whole.
```

Maybe?

dvdplm (approved):
Overall a nice cleanup. Left a few nitpicks.

manul/src/protocol/round_id.rs
fjarri: Python reflexes :) That's an RST format code block.

manul/src/protocol/round_id.rs
fjarri: It's mentioned in the `protocol` module docstring, but I suppose it's worth repeating.

manul/src/protocol/round_id.rs
fjarri: Added a comment.

manul/src/protocol/round_id.rs
fjarri: Good point, removed `ungroup()` in favor of more general `split_group()` (it was only used in one method anyway)

manul/src/protocol/round_id.rs
fjarri: Yeah, probably not a good idea to crash so deep in the call stack. Made this method fallible and added an `expect()` in `Round::transition_info()` impl for `EchoRound`, the error is processed normally in other places.

manul/src/protocol/round_id.rs
fjarri: > Is this correct?
> if false, this round will definitely not return a FinalizeOutcome::Result
> if true, this round may or may not return a FinalizeOutcome::Result

Yes.

> Is it correct to say that it will return a FinalizeOutcome::Result if the protocol was executed correctly, i.e. if there were no errors and this boolean is true, then this round will produce a result?

For CGGMP that's the case, but I can't say for certain it'll always be true. But we don't really need an assertion this strong; this field is used in the chain combinator to let us know in the last round of the first protocol that there may be a message from the first round of the second protocol.

manul/src/protocol/round_id.rs
fjarri: Renamed to `new_linear()` and `new_linear_terminating()`. Not entirely happy with the names still, but there's time until the next release.

manul/src/protocol/round_id.rs
fjarri: Not quite happy with these names either. The state is encapsulated in whatever struct implements `Round`, and "topology" would be applied to the shape of the whole graph, not to a specific node. The pros for the current one is that "state transition" is standard terminology, and it complements "communication info".

manul/src/protocol/round.rs
fjarri: I think that's correct as it is. This variant is used in any round that transitions to another round, it's not specific to combinators.

PR #95, opened 2025-02-20
Proper threshold support

### LINKED ISSUE #11: Support for nodes dropping out during the protocol
Rounds will need to provide a "strategy" describing what nodes they're expecting messages from, and which messages are enough to finalize. Probably returned from `Round::expecting_messages_from` instead of the current `BTreeSet`. Possible options include: all of the nodes, a threshold, maybe several separate sets with their own threshold.

Echo round will need to be mindful of these strategies too when implementing its own `expecting_messages_from` and checking for errors.

### LINKED ISSUE #25: Decouple `protocol` and `session` levels

fjarri:
Fixes #11

Done:
- introduced quorum concept in the API, using it for normal rounds

TODO:
- a method of `IdSet` that would tell us if quorum can never be reached given the already banned nodes
- don't expect messages from a node if it was already banned in the previous round
- support threshold in the echo round, and figure out what `CommunicationInfo` should be like for that
- depending on how the echo round is handled, may be able to close #25?


manul/src/protocol/round.rs
```diff
@@ -19,9 +19,60 @@ use super::{
...
+
+/// Encapsulates the communication info for the main round and the possible echo round.
 #[derive(Debug, Clone)]
 pub struct CommunicationInfo<Id> {
+    /// Communication info for the main part of the round (that is, not considering the echo round).
+    pub main_round: RoundCommunicationInfo<Id>,
+
+    /// The specific way the node participates in the echo round following this round.
+    ///
+    /// `None` means that, if there is an echo round, the message destinations and expected messages senders
```
dvdplm, line 68: Writing `None` here is a bit confusing given it's not an `Option`. Maybe reword this as "An enum describing the way the node participates in the echo round following this round. It can take one of three values, None, SameAsMainRound and Custom, where `None` means that, if there is an echo round, the message destinations and … …."? (Although: what does `SameAsMainRound` even mean then?)

manul/src/protocol/round.rs
```diff
@@ -19,9 +19,60 @@ use super::{
...
+        ids.len() >= self.threshold
+    }
+
+    pub(crate) fn is_quorum_possible(&self, banned_ids: &BTreeSet<Id>) -> bool {
+        let ids = self.ids.intersection(banned_ids).collect::<BTreeSet<_>>();
+        self.ids.len() - ids.len() >= self.threshold
+    }
+}
+
+/// Encapsulates the communication info for the main round and the possible echo round.
```
dvdplm, line 60: I think it'd be better to beef up this doc comment and thin out the lower-level ones a bit, so that the doc reader can learn about the `CommunicationInfo` struct's purpose and usage in one spot.

dvdplm (commented):
Left a few comments and suggestions even though it's a draft.

PR #96, opened 2025-03-04
Upgrade the rand* crates to v0.9

dvdplm:
Upgrade the rand* crates to v0.9 so we can use `manul` with pre-release versionscrypto-primes, crypto-bigint and synedrion.


fjarri:
I think with `manul` 0.2 out, we can get this in for 0.3.

PR #99, opened 2025-04-29
Remove the node from `still_have_not_sent_messages` if it sent an invalid message

fjarri:
In draft, want to think a bit if we can only do this check once before branching out into correct/incorrect message processing.

### DIFF
--- a/manul/src/session/session.rs
+++ b/manul/src/session/session.rs
@@ -611,24 +611,34 @@ where
 
     fn register_unprovable_error(&mut self, from: &SP::Verifier, error: RemoteError) -> Result<(), LocalError> {
         if self.unprovable_errors.insert(from.clone(), error).is_some() {
-            Err(LocalError::new(format!(
+            return Err(LocalError::new(format!(
                 "An unprovable error for {:?} is already registered",
                 from
-            )))
-        } else {
-            Ok(())
+            )));
         }
+        if !self.still_have_not_sent_messages.remove(from) {
+            return Err(LocalError::new(format!(
+                "{:?} was expected to be present in `still_have_not_sent_messages`",
+                from
+            )));
+        }
+        Ok(())
     }
 
     fn register_provable_error(&mut self, from: &SP::Verifier, evidence: Evidence<P, SP>) -> Result<(), LocalError> {
         if self.provable_errors.insert(from.clone(), evidence).is_some() {
-            Err(LocalError::new(format!(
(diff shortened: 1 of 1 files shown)


fjarri:
Actually, I'm having second thoughts. The problem this PR attempts to fix was a problem in `entropy-core` where the nodes sent malformed messages (because of different settings used for serialization/deserialization), and since they didn't pass verification, `still_have_not_sent_messages` in `RoundAccumulator` was not updated (because anyone can send garbage, we can't just count it towards the node it says it's from). So the round could not finalize because it still expected correct messages from all the nodes.

I think this logic is sound, and we can only update `still_have_not_sent_messages` after the message passes verification (as it happens on `master`). `entropy-core` should handle this by imposing a timeout on the session (or perhaps we need to add timeout support in `session::tokio` functions?). @dvdplm, what is your opinion?

ameba23:
> Actually, I'm having second thoughts.

Do you mean second thoughts about adding the check in this PR? I would say any extra checks which can let us know strange things are happening are good, even if they can only happen when broken code (outside of this crate) is used.   

I don't have strong opinions on whether to put a timeout in manul or entropy-core.  We could easily wrap the call to `par_run_session` in a `tokio::time::timeout`. But also happy if manul does it for us.

ameba23:
>  If not, that is, if we can only be sure the message is from a specific node after we verified the signature, then this PR should not go in.

I would say that we can be sure that messages originate from the node specified in the `from` field. Assuming we implemented everything correctly (which i guess is a big if). So i guess this should not be merged. :shrug:  

> The problem with the timeout is that I still want to return the report when the timeout expires, and I'm not sure you can have an externally set timeout in tokio and do that

Then probably it makes sense if the timeout is implemented in manul. Possibly with the desired `Duration` being passed in somewhere.

fjarri:
I think a more flexible choice would be to take a cancellation token, allowing the user to set up the timeout or forced cancellation if needed. See https://github.com/entropyxyz/manul/pull/100

fjarri:
> I would say that we can be sure that messages originate from the node specified in the from field. Assuming we implemented everything correctly (which i guess is a big if).

This is an interesting point actually. `manul` does not assume the messages were previously authenticated because it already has authentication built in. So technically you don't need to do it.

You may still want to encrypt the messages, but the CGGMP scheme specifically does not require it (everything that needs to be encrypted is already encrypted by the scheme itself).

Issue #12, opened 2024-10-12
Timeout hint for the round


fjarri:
In some protocols different nodes need different number of messages to finalize the round. Sometimes none. Some rounds take more time than others, and the protocol level has this information. It would be nice to provide some kind of hint for the user of `Session` to adjust the timeout accordingly.

Note also that it's possible that some subset of nodes generates echo messages, so they will need an echo round, while other nodes switch to the next round right away. The latter nodes may still need to know that an echo round is happening so they have to wait more.

PR #100, opened 2025-04-30
Add cancellation support to `session::tokio` functions

fjarri:
This should allow users to impose an external timeout (or to cancel the session loop manually, e.g. on keyboard interrupt). 

Also fixed some error processing lapses in `par_run_session()` where closing of the incoming message channel (or internal channels) was silently ignored.

The timeout support (#12) is still under consideration. The cancellation takes care of the global timeout, or any other cases where the execution needs to be terminated gracefully, but we may still need some partial cutoff times for specific rounds (not clear at this point what the API would be).


fjarri:
@peg FYI

PR #102, opened 2025-05-12
Example: Dining cryptographers

dvdplm:
Implement [Dining cryptographers problem](https://en.wikipedia.org/wiki/Dining_cryptographers_problem) in `manul`, as an example of a non-trivial protocol.

- Implement the example
- Random documentation tweaks outside the example proper
- A "guide" level document thought to be an intro to `manul`. Probably needs more/better material but it's a start.
 - Where should it live?
 - Should it perhaps be a part of the crate docs?
 - Should I remove it and make into a separate PR?

### Questions
- There are plenty of questions in the code
- How should we run examples? In this PR I suggest using `[[bin]]` in the `manul-examples` cargo file so users can run with `cargo run`, but open to other suggestions.
- Does this example complement the existing ones or does it replace them (in part, in full)?
- Do we need more examples or is this enough for now?


examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,325 @@
...
+#[derive(displaydoc::Display, Debug, Clone, Serialize, Deserialize)]
+/// bla
+pub enum DiningError {
+    /// bla
+    R1Err,
+    /// bla
+    R2Err,
+}
+// TODO: for protocols that do not have "provable errors" we should have a shortcut and/or blanket impls.
+// TODO: Why does ProtocolError require Display?
```
fjarri, line 80: Basically, to show some short version of the error in a debugging/logging scenario, where an actual `Debug` would be too long (because it may contain auxiliary information)

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,325 @@
...
+pub struct DiningCryptographersProtocol;
+#[derive(displaydoc::Display, Debug, Clone, Serialize, Deserialize)]
+/// bla
+pub enum DiningError {
+    /// bla
+    R1Err,
+    /// bla
+    R2Err,
+}
+// TODO: for protocols that do not have "provable errors" we should have a shortcut and/or blanket impls.
```
fjarri, line 79: There is an empty `NoProtocolErrors` that implements `ProtocolError`.

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,325 @@
...
+    ) -> Result<(), ProtocolValidationError> {
+        match self {
+            DiningError::R1Err => Ok(()),
+            DiningError::R2Err => Ok(()),
+        }
+    }
+}
+
+impl<Id> Protocol<Id> for DiningCryptographersProtocol {
+    // TODO: Is this the right place for the final outcome of the protocol?
```
fjarri, line 108: Yep

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,325 @@
...
+    // XOR/¬XOR of the two bits of the diners (one their own cointoss, the other shared with their neighbour).
+    type Result = bool;
+
+    type ProtocolError = DiningError;
+
+    // TODO: Having blanket impls for these methods would cut down on the boilerplate a bit.
```
fjarri, line 114: True, but what would be the default behavior? If it's just ignoring the error, there is a danger of forgetting to implement them. 

That said, I am not happy with these methods. They are far away from the message type, and not very convenient to write and debug. There should be a better way - somehow bundle these checks with the message type itself.

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+// TODO: impl `Round` for `Round1`.
+
+impl Round<Verifier> for Round1 {
+    type Protocol = DiningCryptographersProtocol;
+
+    fn transition_info(&self) -> TransitionInfo {
+        let mut children = BTreeSet::new();
+        children.insert(RoundId::new(2));
+        TransitionInfo {
```
fjarri, line 152: You can use `::new_linear()` constructor here.

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,325 @@
...
+        todo!("Round1 receive message")
+    }
+
+    fn finalize(
+        self: Box<Self>,
+        rng: &mut dyn CryptoRngCore,
+        payloads: BTreeMap<Verifier, Payload>,
+        artifacts: BTreeMap<Verifier, Artifact>,
+    ) -> Result<FinalizeOutcome<Verifier, Self::Protocol>, LocalError> {
+        // Prolly just a NOP in this protocol? Maybe its artifact are the two bits (one private, one shared)?
```
fjarri, line 188: Since you have the second round, it should at least return `Ok(FinalizeOutcome::AnotherRound(BoxedRound::new_dynamic(Round2 { .. })))` (yep, ergonomics is not great here. Can be improved)

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,325 @@
...
+#[derive(Debug, Clone, Deserialize, Serialize, Ord, PartialOrd, Eq, PartialEq)]
+// TODO Need to store the id of the diner and whether they paid or not.
+pub struct Diner(u8);
+impl Diner {
+    fn new(id: u8) -> Self {
+        Self(id)
+    }
+}
+
+// TODO: I don't understand why there needs to be a Signer and a Verifier type. Suspect it's a CGGMP "contamination" and that they are not needed at all for this protocol.
```
fjarri, line 249: It's rather that `manul` assumes you need authentication and evidence generation for your protocol (which you generally would in a production setting). Not needing those leads to unused arguments.

fjarri: You can also use `manul::dev::TestSigner`/`TestVerifier` which do pretty much what you've done here.

dvdplm: Yeah, I tried to do without using `manul::dev` but ended up doing pretty much exactly the same thing.

examples/src/dining_cryptographers.rs
dvdplm: You mean: protocols are so different that there's no reasonable default here? That's fair I guess. OTOH there's a limit to the amnount of hand-holding a library can provide beyond "make it easy to do the right thing". Tricky to know, but I guess the it's an indication that writing examples is useful to see where the sore spots are.

examples/src/dining_cryptographers.rs
dvdplm: I'd argue that it's prooobably not our business (as library authors) to decide if a `Debug` output is "too long" and imposing `Display` as a stand-in for `Debug` is a bit odd.

manul/src/session/session.rs
```diff
@@ -642,6 +642,9 @@ where
         }
     }
 
+    // Add a processed artifact to the accumulator.
+    // Returns an error if the artifact was already present.
+    // Called by [`Session::add_artifact`].
```
fjarri, line 647: This type of comment tends to get out of sync unexpectedly :)

manul/src/protocol/round.rs
```diff
@@ -258,6 +258,7 @@ impl<Id> ProtocolError<Id> for NoProtocolErrors {
 }
 
 /// Message payload created in [`Round::receive_message`].
+// TODO: Clarify the purpose of this type, in particular vis-a-vis [`Artifact`].
```
fjarri, line 261: So, "payload" is the result of processing an incoming message, while "artifact" is the result of creating a message locally (besides the message itself). I don't particularly like "payload" myself, it's too general.

manul/src/protocol/round.rs
```diff
@@ -280,14 +281,21 @@ impl Payload {
...
+        })?))
     }
 }
 
 /// Associated data created alongside a message in [`Round::make_direct_message`].
+// TODO: Is the artifact sent to other participants (public) or is it private data, available only to the party that
+// produced it?
+// Seems like Artifacts are private, and never hit the wire. Artifacts are collected and "delivered"
+// in the `Round::finalize()` method in the form of a BTreeMap, where the key is the *destination* of the direct message.
+// Is the right primitive to store state in? No, I think not.
```
fjarri, line 298: Yes, an artifact is usually some secret that gets generated when creating a message, and is needed in subsequent rounds. What would be the alternative? We can just get rid of it completely and have nodes generate all the required secrets in the round constructor, but that will mean some non-parallelized work, and also some extra work if we ever have threshold support.

manul/src/dev/run_sync.rs
```diff
@@ -173,6 +193,7 @@ where
         let session = Session::<_, SP>::new(rng, session_id.clone(), signer, entry_point)?;
         let mut accum = session.make_accumulator();
 
+        // TODO: why are we doing this here AND inside `propagate`?
```
fjarri, line 196: `propagate()` is called twice, one time we do need to "pre-seed" the messages, the other we don't. I remember being irked by this as well, but I couldn't make the logic work otherwise. If you see how to fix it, would be nice.

manul/src/dev/run_sync.rs
```diff
@@ -134,10 +141,23 @@ where
...
+            CanFinalize::Never => {
+                trace!(
+                    "[{:?}] Can't ever finalize. Terminating session due to errors",
+                    session.verifier()
+                );
+                break State::Finished(session.terminate_due_to_errors(accum)?);
+            }
         }
-
+        // TODO: didn't we just make a bunch of calls to make_message and processed the artifact?
```
fjarri, line 160: Could you elaborate? You mean the API is too awkward?

examples/GUIDE.md
```diff
@@ -0,0 +1,121 @@
+# User's Guide: Getting started building protocols with `manul`
```
fjarri, line 1: I wonder if this can be included in the main documentation somehow.

examples/GUIDE.md
```diff
@@ -0,0 +1,121 @@
...
+impl<Id> Protocol<Id> for MyProtocol {
+    type Result = SomeResultType;
+    type Error = SomeErrorType;
+    // ... other required trait methods ...
+}
+```
+
+Key aspects:
+
+- **`type Result`**: This associated type defines the final output of a successful protocol execution. In the case of the Dining Cryptographers Problem, it's a thruple of bools representing each cryptographer's perspective on the outcome.
```
fjarri, line 26: *tuple

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,487 @@
+//! Dining Cryptographers Protocol, see https://en.wikipedia.org/wiki/Dining_cryptographers_problem
```
ameba23, line 1: Great explanation

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,487 @@
...
+// combining sub-protocols together.
+// For this simple protocol it's mostly boiler plate, with the exception of the `Result` associated type: it defines
+// what the final outcome of a protocol run actually is.
+#[derive(Debug)]
+pub struct DiningCryptographersProtocol;
+
+// TODO: I guess the protocol is a place where one could store arbitrary state/context/config data as well? We don't do
+// that much in `synedrion` but would it be a good idea to do so?
+impl<Id> Protocol<Id> for DiningCryptographersProtocol {
+    // XOR/¬XOR of the two bits of each of the three diners (one is their own cointoss, the other shared with their
```
ameba23, line 84: Ok so this is fixed at 3 parties, and the result it the XOR result from diner A, diner B and diner C?

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,487 @@
...
+    type Result = (bool, bool, bool);
+
+    type ProtocolError = NoProtocolErrors;
+
+    fn verify_direct_message_is_invalid(
+        format: &BoxedFormat,
+        round_id: &RoundId,
+        message: &DirectMessage,
+    ) -> Result<(), MessageValidationError> {
+        Ok(())
```
ameba23, line 95: These relate to serialization errors, right?

Not related to this PR, but i find the documentation for these confusing:
https://github.com/entropyxyz/manul/blob/40c867b85576ea3143a5ff26bfdccb347d2c70dd/manul/src/protocol/round.rs#L76
Should it not be 'can' rather than 'cannot'?

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,487 @@
...
+    fn verify_normal_broadcast_is_invalid(
+        format: &BoxedFormat,
+        round_id: &RoundId,
+        message: &NormalBroadcast,
+    ) -> Result<(), MessageValidationError> {
+        Ok(())
+    }
+}
+
+// HELP: Keeping state in the round like this is a bit weird. Is this the intended way to do it? It would have been
```
ameba23, line 115: :shrug: - i guess so - would you say this stuff was also state? https://github.com/entropyxyz/synedrion/blob/1b597127d2667b4fd4136ca31c7a462a0cabe35a/src/protocols/interactive_signing.rs#L1093

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,487 @@
...
+        let round = BoxedRound::new_dynamic(round);
+        Ok(round)
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, Ord, PartialOrd, Eq, PartialEq)]
+// TODO: this type is not used at all. It'd be the natural place for things like "paid for dinner" and other similar
+// state, but I don't see how to access a `Diner` from inside the protocol, without doing nasty stuff like having a
+// global "state" thingy (standing in for a database).
+pub struct Diner {
```
ameba23, line 385: Why do we have both `Diner` and `DinerId`?

examples/src/dining_cryptographers.rs
fjarri: Originally it was a separate `description()` method, perhaps it makes sense to bring it back.

examples/src/dining_cryptographers.rs
dvdplm: It is indeed fixed and limted to exactly 3 parties. There are ways to generlize to any number of parties, apparently called "DC nets" (probably to make it as hard as possible to google).

examples/src/dining_cryptographers.rs
dvdplm: Opened https://github.com/entropyxyz/manul/issues/105

examples/src/dining_cryptographers.rs
dvdplm: Yeah, the `Context` struct is state. As a developer using `manul` I would like to know what the best practise is for local state, so I don't end up mixing up what I actually need to send around to other participants with local state.

@fjarri Do you have any input here? Is this something we can "solve" in code you reckon? Or document better?

examples/GUIDE.md
```diff
@@ -0,0 +1,121 @@
...
+
+This guide provides a step-by-step explanation of how to build protocols using `manul`. See the examples in this folder for concrete examples.
+
+Building a protocol with `manul` involves defining the protocol's overall structure, breaking it down into rounds with specific communication and logic, defining the messages exchanged, providing an entry point to initialize the protocol, configuring session parameters for cryptography and data handling, and finally using the provided utilities to execute the protocol (or write your own).
+
+Refer to the crate documentation for more details and advanced topics.
+
+## 1. Define Your Protocol Structure (`Protocol`)
+
+Start by defining a struct to represent your protocol. This struct will implement the `Protocol` trait, serving as a container for the protocol's core properties:
```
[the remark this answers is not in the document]
fjarri: If it's in the main docs all this stuff would be easier to reference

examples/src/dining_cryptographers.rs
fjarri: You can extract the state in its own type, then you'll have `Round1 { state: State }` and `Round2 { state: State, paid: bool }`. `Diner` is something related to the protocol execution, and the protocol implementation (rounds etc) cannot know anything about it, since it may easily be somewhere in the dependent crate.

manul/src/session/message.rs
```diff
@@ -199,7 +199,7 @@ impl<M> VerifiedMessagePart<M> {
 }
 
 /// A signed message destined for another node.
-#[derive(Clone, Debug, Serialize, Deserialize)]
```
fjarri, line 202: Seems like it's reverting #98

manul/src/dev/run_sync.rs
dvdplm: Maybe we can fix it, not sure yet.

examples/GUIDE.md
dvdplm: LOL, yep, missed this one.

examples/Cargo.toml
```diff
@@ -9,17 +9,23 @@ repository = "https://github.com/entropyxyz/manul/examples"
...
 
 [dev-dependencies]
 tokio = { version = "1", features = ["rt", "sync", "time", "macros"] }
 rand = "0.8"
 digest = "0.10"
 manul = { path = "../manul", features = ["dev", "tokio"] }
 test-log = { version = "0.2", features = ["trace", "color"] }
+
+[[bin]]
+name = "dining-cryptographers"
```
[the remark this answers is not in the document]
dvdplm: I started that way and then ran into some snag I now have forgotten. :/

manul/src/session/message.rs
dvdplm: :/

manul/src/dev/run_sync.rs
```diff
@@ -228,7 +259,6 @@ where
         };
         reports.insert(verifier, report);
     }
-
```
fjarri, line 231: Just out of curiosity, I noticed removed empty lines here and there; is it:
- an intentional thing,
- something your editor does on save,
- an artifact of having some debug code there that was later removed?

dvdplm: Probably a mix of rustfmt'ing on save a + debug code that was removed; definitely not intentional! I'll pay more attention to this to cut down on the noise.

manul/src/protocol/round.rs
```diff
@@ -303,27 +314,29 @@ impl Artifact {
...
+            ))
+        })?))
     }
 }
 
-/// A round that initiates a protocol.
-///
-/// This is a round that can be created directly;
-/// all the others are only reachable throud [`Round::finalize`] by the execution layer.
+/// A round that initiates a protocol and defines how execution begins. It is the only round that can be created outside
```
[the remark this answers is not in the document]
dvdplm: Good point, I actually did not know that, ty.

manul/src/dev/run_sync.rs
dvdplm: Undid these removals.

examples/src/dining_cryptographers.rs
```diff
@@ -0,0 +1,471 @@
...
+//!    Each diner broadcasts one bit: the XOR (if they didn't pay) or ¬XOR (if they paid) of their two bits.
+//!    Outcome: each diner has a set of bits, one for each diner.
+//! Post:
+//!    Everyone XORs all bits together and interprets the result: 0 => NSA paid; 1 => one of the diners paid.
+//!
+//! ## Running the example
+//!
+//! Run the example with `RUST_LOG=trace cargo run dining-cryptographers`.
+
+#![allow(dead_code, unused, unused_imports)]
```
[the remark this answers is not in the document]
dvdplm: When fixing this I ran into a bit of an issue: `rustc` throws warnings about `main` not being used, even when `cargo run`-ing it. I think it's because `manul-examples` is a library crate and `src/dining-cryptographers.rs` is a module in `lib.rs` and so the compiler sees `main()` like any other library item.

I moved `dining-cryptographers.rs` outside `src`, so it's not actually part of the library API of `manul-examples` and "just" a stand-alone binary as configured in `manul-examples/Cargo.toml`. 

It looks a bit odd/unusual but:

- no warnings
- not part of the "API" of the `manul-examples` crate
- runs from the workspace root with `cargo run -p manul-examples`

It's still a bit unsatisfying that some examples run as tests, others as binary executables and none are "normal" cargo examples (i.e. examples that would live inside `manul/examples` and run with `cargo run --example NAMEHERE`).

@fjarri We'll want to make a decision on where we want our examples to be so we give users a consistent way. Do you agree? Follow-up PR or in here?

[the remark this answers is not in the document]
fjarri: The reason why I initially created the example crate as a library is that it is the way protocols are supposed to be implemented - they don't execute anything, just provide the stuff to be executed in a `Session` elsewhere. So the nonsensical `SimpleProtocol` was organized as a library + tests. Do you think that's too complicated for an example?

fjarri:
The code looks good to me, but I'm still not convinced about the directory layout. It looks weird, especially next to the other example which is a library. Are you sure this makes it easier to understand as opposed to it being in `src/`? It would still be a single file, just `main()` would be a test.

fjarri:
Just realized there's #115. I guess this'll go in then, but I do hope we clean it up soon (famous last words, I know).

fjarri (approved):

Issue #65, opened 2024-11-05
Typed `Round` trait


fjarri:
The current dynamic rounds are useful for writing combinators and tests, but we need to explore more static rounds (that would still be castable to `BoxedRound` and compatible with the rest of the machinery). Something that has associated types for the messages, payloads, and artifacts so that they don't have to be manually serialized/deserialized.

The problem here is correctly recognisinig rounds that don't send some type of messages, or don't create artifacts. We can use `()` in the corresponding associated type to indicate that.

dvdplm:
What do you mean by "more static rounds" here?

fjarri:
So the idea was to have something like
```rust
trait StaticRound<Id> {
    type DirectMessage;
    ...
    fn make_direct_message(...) -> Result<DirectMessage, LocalError>;
    ...
    fn finalize(direct_messages: BTreeMap<Id, Self::DirectMessage>, ...) -> ...;
}
```
and handle serialization/deserialization (and failures thereof) internally, without having to write them down explicitly in each `Round` implementation.

The problem is handling the case of no messages; we would want something like
```
type DirectMessage = NoMessage;
```
and the default implementation of `make_direct_message()` that will return it. But that seems to be impossible to do without trait specialization (which is currently unstable). 

The other option is to do what `synedrion` did before the generic round stuff was extracted to `manul`, and use macros:
```
impl StaticRound for Round1 {
    no_direct_messages!()
 
    // Expands to 
    // type DirectMessage = NoMessage;
    // fn make_direct_message(...) -> Result<DirectMessage, LocalError> { Ok(NoMessage) }
}
```

fjarri:
Actually, there's a hacky serialize/deserialize way around it. I am going to make a PR to illustrate.

PR #112, opened 2025-05-26
Rename `BoxedRound::new_dynamic` to just `new`

dvdplm:
Given that there are no other constructors available on `BoxedRound` I think it's better to name it just `new`.


fjarri:
Yeah, perhaps. It was a preparation for https://github.com/entropyxyz/manul/issues/65 which I'm not sure is coming

PR #116, opened 2025-06-05
Re-seed the RNG before spawning

### LINKED ISSUE #104: Re-seed the RNG before passing it to a spawned task
This mostly applies to creating direct messages. We may be passing the same RNG state to every task, which is a security issue.

dvdplm:
Any CSPRNG that is `Clone` and has internal state (e.g. all RNGs that are`SeedableRng`) is hazardous when used in spawned tokio tasks: each task will have an exact copy of the parent state and each use the same randomness (unless the CSPRNG author made sure this doesn't happen, which is not mandated by the type system, so we cannot know).

One commonly used CSPRNG is the `OsRng`, which is `Clone` but not `SeedablRng`. This particular CSPRNG is actually safe to clone and use in spawned tasks, because its state is fetched from the OS each time it is used (at the price of a performance hit).

Our parallel protocol runner requires `impl 'static + Clone + CryptoRngCore + Send`, which works well with `OsRng` and most other CSPRNGs, but we cannot re-seed it without also requiring `SeedableRng` (and thus forbid `OsRng`).

To square this circle this PR suggests we make an opinionated choice of creating new `ChaCha20` CSPRNGs for each task, seeded with the provided `impl CryptoRngCore`. It is a safe, performant and well-tested RNG that is unlikely to cause issues in the future.
 

Fixes #104

### DIFF
--- a/manul/Cargo.toml
+++ b/manul/Cargo.toml
@@ -17,6 +17,7 @@ serde-encoded-bytes = { version = "0.2", default-features = false, features = ["
 digest = { version = "0.10", default-features = false }
 signature = { version = "2", default-features = false, features = ["digest", "rand_core"] }
 rand_core = { version = "0.6.4", default-features = false }
+rand_chacha = { version = "0.3", default-features = false }
 tracing = { version = "0.1", default-features = false }
 displaydoc = { version = "0.2", default-features = false }
 derive-where = "1"
--- a/manul/Cargo.toml
+++ b/manul/Cargo.toml
@@ -17,7 +17,7 @@ serde-encoded-bytes = { version = "0.2", default-features = false, features = ["
 digest = { version = "0.10", default-features = false }
 signature = { version = "2", default-features = false, features = ["digest", "rand_core"] }
 rand_core = { version = "0.6.4", default-features = false }
-rand_chacha = { version = "0.3", default-features = false }
+
 tracing = { version = "0.1", default-features = false }
 displaydoc = { version = "0.2", default-features = false }
 derive-where = "1"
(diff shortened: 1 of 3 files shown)


manul/Cargo.toml
```diff
@@ -17,6 +17,7 @@ serde-encoded-bytes = { version = "0.2", default-features = false, features = ["
 digest = { version = "0.10", default-features = false }
 signature = { version = "2", default-features = false, features = ["digest", "rand_core"] }
 rand_core = { version = "0.6.4", default-features = false }
+rand_chacha = { version = "0.3", default-features = false }
```
fjarri, line 20: If it's only used in `tokio`, it should only be included with that feature.

manul/src/session/tokio.rs
```diff
@@ -234,14 +235,13 @@ where
         let destinations = session.message_destinations();
         let mut message_creation_tasks = Vec::new();
         for destination in destinations {
-            let rng = rng.clone();
             let session = session.clone();
             let my_id = my_id.clone();
             let outgoing_tx = outgoing_tx.clone();
             let destination = destination.clone();
+            let mut task_rng = ChaCha20Rng::from_rng(rng.clone()).map_err(|_| LocalError::new("Can't fork the RNG"))?;
```
fjarri, line 242: I think a short comment would be useful here, explaining why we're doing this.

manul/src/session/tokio.rs
```diff
@@ -234,14 +235,13 @@ where
         let destinations = session.message_destinations();
         let mut message_creation_tasks = Vec::new();
         for destination in destinations {
-            let rng = rng.clone();
```
fjarri, line 237: This should make the `Clone` bound unnecessary?

